PR #1, opened 2015-03-15
[NOT READY] Import and reorganize account creation/filtering

chriskuehl:
This imports a lot of code from create, which was written collectively
by OCF staffers, but especially:

- Nader Morshed <morshed@ocf.berkeley.edu> (primary author)
- Kenny Do <kedo@ocf.berkeley.edu> (account filtering)
- Dara Adib <daradib@ocf.berkeley.edu> (realname checking, minor fixes)

Much of the code has been reorganized or slightly rewritten in
preparation of switching to single-account-transaction account creation
(rather than batch processing of several accounts at once).

Since we also no longer use the 'mid' stage of account creation (where
accounts have been approved, but wait 24 hours to be created), that code
is also removed.

Currently, this imports most of the account filtering, but not all of
the account creation. The main things still remaining are:

* LDAP entry creation
* Kerberos principal creation

There are some smaller things still remaining as well:

* Getting max UID

### DIFF
--- a/ocflib/account/creation.py
+++ b/ocflib/account/creation.py
@@ -1,0 +1,245 @@
+import difflib
+import itertools
+import math
+import os.path
+import re
+import subprocess
+import sys
+
+import ocflib.account.search as search
+import ocflib.account.utils as utils
+import ocflib.constants as constants
+import ocflib.misc.mail as mail
+import ocflib.validators as validators
+
+
+def create_home_dir(user):
+    """Create home directory for user. Makes a directory with appropriate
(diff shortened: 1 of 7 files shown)


chriskuehl:
To clarify, this imports most of the core (functionality) of create, but not the CLI bits (those are pretty small). The plan is for those to stay, we just make it call ocflib to do actual work.

Then at some point when we're ready to move to transnational account creation, we can get rid of the CLI and replace it with whatever processes accounts (not sure on this yet, maybe we want to use some kind of task queue or something).

PR #3, opened 2015-09-24
Cache known UID in module

daradib:
We can cache the known UID with a slight hack and prime the cache
on import. If accounts are deleted, UIDs could be skipped.
Cache is not shared across multiple processes.

### DIFF
--- a/ocflib/account/creation.py
+++ b/ocflib/account/creation.py
@@ -75,19 +75,20 @@ def create_account(request, creds, report_status):
 
 
 def _get_first_available_uid():
-    """Returns the first available UID number.
+    """Return the first available UID number.
 
     Searches our entire People ou in order to find it. It seems like there
     should be a better way to do this, but quick searches don't show any.
 
     We hard-code a value we know has already been reached and only select
-    entries greater than that for performance.
+    entries greater than that for performance. We then use a module-level
+    dict to cache our output for the next function call.
     """
-    KNOWN_MIN = 35000
+    min_uid = _cache['known_uid']
     with ldap_ocf() as c:
         c.search(
             constants.OCF_LDAP_PEOPLE,
-            '(uidNumber>={KNOWN_MIN})'.format(KNOWN_MIN=KNOWN_MIN),
+            '(uidNumber>={KNOWN_MIN})'.format(KNOWN_MIN=min_uid),
             attributes=['uidNumber'],
         )
         uids = [int(entry['attributes']['uidNumber'][0]) for entry in c.response]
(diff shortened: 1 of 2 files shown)


chriskuehl:
Maybe we should just do something like suggested here?
http://www.perlmonks.org/?node_id=457205

nattofriends:
lol

2015-09-24 12:35 GMT-07:00 Chris Kuehl <notifications@github.com>:

chriskuehl:
Alternatively, supposedly uids can go to ~2^31-2. Maybe we just give up on sequential uids and randomly pick them in some large range (with a check to avoid duplicates, unlikely as they may be)?

This is almost a serious suggestion. Does anyone actually care about sequential UIDs? We can add a new field in LDAP (and should anyway) for account creation date (instead of estimating by UID). I think we have enough data to backfill that attr for most existing accounts.

nattofriends:
No, that sound really strange.

chriskuehl:
> No, that sound really strange.

sounds strange, but probably more reliable and faster than other methods (short of hard-coding a constant in the code like we are now)

daradib:
@chriskuehl Let's keep things simple. Would suggest merging this or storing highest UID in an LDAP entry (and perhaps also using it for locking).

chriskuehl:
Random UIDs are simpler than complicated caching or an LDAP entry that might become inconsistent with the actual data in LDAP.

nattofriends:
I don't enjoy this. 

How about switching the OpenLDAP backing to MySQL and then directly querying that?

chriskuehl:
@nattofriends that would make me happier as well

nattofriends:
To be clear, "this" was referring to random uids, not caching.

daradib:
I like caching, LDAP record, and MySQL. IMHO, random (U)UID is a neat solution for a problem we don't have. We need unique usernames. We don't/can't create accounts in parallel. And perhaps some legacy application uses 16-bit UIDs.

chriskuehl:
Unless one of our requirements is sequential UIDs (is it?), random UIDs are more robust and easier to implement than anything else being suggested, and satisfy all our requirements. Limiting to 16 bits is not an issue.

daradib:
Do we gain anything? Do we lose anything? Is it easier to implement? OpenLDAP isn't an eventually-consistent datastore. Collisions are unlikely with 32-bit integers but not as practically impossible as 128-bit UUIDs. So you'd still want to check that the UID doesn't exist. I'm not sold, but if you really think so, go ahead.

chriskuehl:
```python
uid = None
while not uid or uid_exists(uid):
    uid = random.randint(100000, 999999)
return uid
```

the implementation is pretty simple...

still not sure I like it better, but I don't see why it's a bad idea.

chriskuehl:
let's just merge this for now and probably do the MySQL thing (rt#3956)

chriskuehl:
thanks for the patch!

PR #4, opened 2015-10-08
Remove middle names shown on staff hours

jvperrin:
This pull request removes the middle names currently shown on the [staff hours page](https://ocf.berkeley.edu/staff-hours). That's about it!

### DIFF
--- a/ocflib/lab/staff_hours.py
+++ b/ocflib/lab/staff_hours.py
@@ -57,7 +57,7 @@ def get_staff_hours():
             staff=[
                 Staffer(
                     user_name=attrs['uid'][0],
-                    real_name=remove_middle_names(attrs['cn'][0]),
+                    real_name=_remove_middle_names(attrs['cn'][0]),
                     position=position(attrs['uid'][0]),
                 ) for attrs in map(user_attrs, hour['staff'])
             ],
@@ -66,9 +66,9 @@ def get_staff_hours():
     ]
 
 
-def remove_middle_names(name):
-    names = name.split(" ")
-    return names[0] + " " + names[-1]
+def _remove_middle_names(name):
+    names = name.split(' ')
+    return names[0] + ' ' + names[-1]
 
 
 def get_staff_hours_soonest_first():
--- a/ocflib/lab/staff_hours.py
+++ b/ocflib/lab/staff_hours.py
@@ -57,13 +57,18 @@ def get_staff_hours():
             staff=[
                 Staffer(
                     user_name=attrs['uid'][0],
-                    real_name=attrs['cn'][0],
+                    real_name=remove_middle_names(attrs['cn'][0]),
                     position=position(attrs['uid'][0]),
                 ) for attrs in map(user_attrs, hour['staff'])
             ],
             cancelled=hour['cancelled'],
         ) for hour in staff_hours['staff-hours']
     ]
+
+
+def remove_middle_names(name):
+    names = name.split(" ")
+    return names[0] + " " + names[-1]
 
 
 def get_staff_hours_soonest_first():


matthew-mcallister:
Would we rather remove the names from LDAP?

nattofriends:
I would rather they be "corrected" in LDAP

jvperrin:
That would be an option, I was mostly just following the existing behavior to remove middle names that appears in the [old staff hours file](https://github.com/ocf/website/blob/master/cgi-bin/staff_hours.cgi#L102). How did the middle names get into LDAP in the first place?

matthew-mcallister:
We get the names from the University directory, which is notorious for
having everyone's full name in ALL CAPS.

chriskuehl:
@jvperrin they're imported from public CalNet data, e.g.

```
ckuehl@supernova:~$ ldapsearch -H ldap://ldap.berkeley.edu -b ou=People,dc=Berkeley,dc=EDU -x uid=1101587
dn: uid=1101587,ou=people,dc=berkeley,dc=edu
objectClass: ucEduPerson
objectClass: person
objectClass: inetorgperson
objectClass: organizationalperson
objectClass: berkeleyEduPerson
objectClass: top
objectClass: eduPerson
o: University of California, Berkeley
ou: people
berkeleyEduModDate: 20151005150855Z
mail: jvperrin@berkeley.edu
berkeleyEduAffiliations: STUDENT-TYPE-REGISTERED
givenName: JASON VENESS
displayName: JASON VENESS PERRIN
sn: PERRIN
cn: PERRIN, JASON VENESS
uid: 1101587
```

jvperrin:
@chriskuehl @matthew-mcallister Ok, good to know. So we would just filter the names when they are imported into LDAP instead then?

matthew-mcallister:
In the past we've just manually updated the names of staff. I don't think
we care about the names of every user.

@chriskuehl <https://github.com/chriskuehl> @matthew-mcallister
<https://github.com/matthew-mcallister> Ok, good to know. So we would just
filter the names when they are imported into LDAP instead then?

—
Reply to this email directly or view it on GitHub
<https://github.com/ocf/ocflib/pull/4#issuecomment-146449998>.

chriskuehl:
The logic is kind of complicated already: https://github.com/ocf/ocflib/blob/master/ocflib/ucb/directory.py#L13

We have to be a little careful with automatic stripping since we might lose something important. In the past we've just changed names in LDAP of staff members so they appear reasonable on the staff hours page.

I don't really feel strongly one way or another.

ocflib/lab/staff_hours.py
```diff
@@ -57,6 +57,11 @@ def get_staff_hours():
     ]
 
 
+def remove_middle_names(name):
+    names = name.split(" ")
+    return names[0] + " " + names[-1]
```
chriskuehl: nit: [pre-commit](http://pre-commit.com/) will force you to use single quotes instead of double quotes, so this commit is actually failing the tests)

it's pretty arbitrary (in Python, there's no important difference, even with control sequences like `\n`) but it helps keep things consistent. Similarly it does things like sort imports.

I'd recommend running `pre-commit install` when in the repo which will install pre-commit hooks so that the tests run when you make a commit. (If you're on your staff VM, you might need to `apt-get install pre-commit` first.)

You can also run `tox` to run all the tests, including pre-commit. Will make a PR to improve that (imo `make test` should just do that, similar to ocfweb).

nattofriends: >pre-commit will force you to use single quotes

disgusting

2015-10-08 16:06 GMT-07:00 Chris Kuehl <notifications@github.com>:

jvperrin: I also had to install dch through the `devscripts` package (along with `pre-commit` of course) to get the tests to pass properly.

chriskuehl: ah yeah, that's because of the autoversion target that generates the Debian changelog

take a look at #5 -- it actually fixes that as well (tox no longer requires autoversion)

jvperrin: Oh ok, cool, already fixed before I even knew it was an issue! :open_mouth:

daradib:
>The logic is kind of complicated already:
>https://github.com/ocf/ocflib/blob/master/ocflib/ucb/directory.py#L13

Back in ~2011, I think I asked @kennydo to concatenate the longest first and last names in CalNet LDAP, since that a) seemed to match names on Cal ID b) gave us the most information, like middle names, to distinguish between people with similar names. Was more important when we didn't have self-service online password changes. Now we can always clarify with CalNet LDAP, except for alumni records, which are private.

chriskuehl:
if you want to ship this, i'm still in favor, will need to fix comments though

fixing in ldap is another option of course

jvperrin:
It still needs to be fixed in LDAP for the capitalization of names unfortunately. What do you mean by the comments?

chriskuehl:
s/comments/conflicts/

lol whoops

jvperrin:
Oh ok, that makes more sense, I'll do that then.

PR #6, opened 2015-10-29
Add CalLink API search by group name

matthew-mcallister:


ocflib/ucb/groups.py
```diff
@@ -18,12 +18,48 @@
...
+        'category': '',
+    }, _API['SERVICE']['ORGS'], parser)
+
+
+def group_name_by_oid(oid):
+    """Return the name and OCF account(s) of a group.
+
+    >>> group_name_by_oid(46187)
+    {'name': 'Open Computing Facility', accounts: ['decal', 'linux']}
+    """
```
chriskuehl: confusing function name since it returns more than just the group's name.
maybe `group_by_oid` would be better?

ocflib/ucb/groups.py
```diff
@@ -18,12 +18,48 @@
 }
 
 
-# TODO: add method(s) for using CalLinkOrganizations service
+def groups_by_name(name, oid=''):
+    """Return groups by matching name.
+
+    >>> groups_by_name("facility")
+    {46187: {'name': 'Open Computing Facility', accounts: ['decal', 'linux']}}
+    """
```
chriskuehl: maybe make this more generic? it's weird that it takes a required name (but sometimes you pass an empty string) and an optional oid

i'd suggest:
* change signature to `def list_groups(name=None, oid=None, status=None, type=None, category=None):`
* always deal with oids as integers to avoid confusion (so you'll need to `str(...)` it); we already `int` it when returning
* change return statement to
```python
    return _get_osl({
        'name': name or '',
        'organizationId': str(oid) if oid else '',
        'status': status or '',
        'type': type or '',
        'category': category or '',
    }, _API['SERVICE']['ORGS'], parser)
```

then it solves basically any future need we might have

chriskuehl:
looks pretty great overall

ocflib/ucb/groups.py
matthew-mcallister: This was by analogy with `groups_by_student_signat` and `groups_by_student_signat_all`, but I agree.

ocflib/ucb/groups.py
matthew-mcallister: Agreed. I don't know what `status`, `type`, or `category` are for so I wasn't sure how useful they'd be in the function signature.

matthew-mcallister:
I'm still making changes to this as I'm rewriting `signat`. Stay tuned.

matthew-mcallister:
Signat is done, waiting further evaluation here.

ocflib/ucb/groups.py
```diff
@@ -18,12 +18,48 @@
 }
 
 
-# TODO: add method(s) for using CalLinkOrganizations service
+def list_groups(name=None, oid=None, status=None, type=None, category=None):
+    """Return groups by a general CalLink search.
+
+    >>> list_groups("facility")
+    {46187: {'name': 'Open Computing Facility', accounts: ['decal', 'linux']}}
```
chriskuehl: doctest is wrong now
(ok, technically it works, but only because it will fill the first kwarg. should update the example to `name='facility'`)

chriskuehl:
shipit assuming `make test` passes \o/

matthew-mcallister:
Fixing and shipping.

PR #13, opened 2016-01-27
Use datetime.time for open and close hours

matthew-mcallister:
Also set Tuesday opening to 9:10 am. This will probably break ocfweb.

### DIFF
--- a/ocflib/lab/hours.py
+++ b/ocflib/lab/hours.py
@@ -17,12 +17,13 @@ from collections import defaultdict
 from collections import namedtuple
 from datetime import date
 from datetime import datetime
+from datetime import time
 
 
 class Hour(namedtuple('Hours', ['open', 'close'])):
 
     def __contains__(self, when):
-        return self.open <= when.hour < self.close
+        return self.open <= when.time() < self.close
 
 
 class Day(namedtuple('Day', ['date', 'weekday', 'holiday', 'hours'])):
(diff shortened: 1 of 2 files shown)


ocflib/lab/hours.py
```diff
@@ -17,12 +17,21 @@
...
 from datetime import datetime
+from datetime import time
 
 
-class Hour(namedtuple('Hours', ['open', 'close'])):
+def Hour(opn, cls):
+    if type(opn) == int:
+        opn = time(opn)
+    if type(cls) == int:
+        cls = time(cls)
```
chriskuehl: i'm not a huge fan of this, adding `time(...)` is small enough that i don't think it's necessary to add a confusing constructor that accepts two different types.

i'd probably just do

```python
     Day.FRIDAY: [Hour(time(9), time(20))],
      Day.SATURDAY: [Hour(time(11), time(19))],
      Day.SUNDAY: [Hour(time(11), time(19))],
```

i don't really think it's that much harder to read

if you want to keep is this way,

1. use `is` for comparing singletons like types, e.g. `if type(opn) is int:`
2. probably better to use `isinstance` anyway, e.g. `if isinstance(opn, int):`

matthew-mcallister: Oh, well I agree the constructor is not very good. I just didn't want to have to refactor every repo using ocflib. But if the constructor is only called in this file (and I don't see where else it would be anyways) that's much simpler.

matthew-mcallister:
Nuked original commit, awaiting approval.

ocflib/lab/hours.py
chriskuehl: yeah, good point. but i'm pretty sure it's only called here so it should be safe? hopefully...

chriskuehl:
sorry for the slow review, this looks good to me

verifying quickly that ocfweb tests fail to give you a chance to merge the other PR

chriskuehl:
yup, good to merge.
ocfweb will fail but that's what we want.

PR #15, opened 2016-02-11
Reserve usernames of some former staffers whose accounts have been lost

kpengboy:
So we won't have more problems with random future members being confused with
old board members, or more significantly User_Info entries incorrectly
labeling new members as such


chriskuehl:
shipit after flake8 fix

kpengboy:
Revised and committed.

PR #20, opened 2016-03-02
Add quota schema and simple functions for manipulating it

chriskuehl:
You can log in to mysql with anonymous (no password) to see the
`printed` view which forms the core of the accounting.

You can log in with the ocfprinting user (in passwords file) to see the
full jobs and refunds tables, plus some additional views.

The functions like get_quota work already (try accounts ckuehl, asdf,
hjkl, pubstaff, and some you make up to see how it handles non-existent
users).

### DIFF
--- a/ocflib/printing/quota.py
+++ b/ocflib/printing/quota.py
@@ -170,9 +170,9 @@ def get_schema():
                 DETERMINISTIC
                 BEGIN
                 IF MONTH(d) > 8 THEN
-                    RETURN MAKEDATE(YEAR(d), 213);
+                    RETURN MAKEDATE(YEAR(d), 213);  -- roughly august 1st
                 ELSE
-                        RETURN MAKEDATE(YEAR(d), 1);
+                    RETURN MAKEDATE(YEAR(d), 1);
                 END IF;
             END$$
         DELIMITER ;
--- a/ocflib/printing/quota.py
+++ b/ocflib/printing/quota.py
@@ -1,6 +1,5 @@
 from collections import namedtuple
 from datetime import datetime
-from textwrap import dedent
 
 import pymysql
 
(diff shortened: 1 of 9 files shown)


chriskuehl:
@jvperrin @matthew-mcallister

chriskuehl:
@kpengboy

ocflib/printing/quota.py
```diff
@@ -0,0 +1,236 @@
...
+        autocommit=True,
+    )
+
+
+def get_schema():
+    """Return the printing schema.
+
+    Normally, this never needs to be called and just exists to document the
+    existing schema.
+    """
```
chriskuehl: this kinda sucks but we don't want to do migrations and such

ocflib/printing/quota.py
```diff
@@ -0,0 +1,236 @@
...
+UserQuota = namedtuple('UserQuota', (
+    'user',
+    'daily',
+    'semesterly',
+))
+
+
+Job = namedtuple('Job', (
+    'user',
+    'time',
```
jvperrin: Should the time fields here and in the Refund `namedtuple` have a default value consisting of whenever they are created? It seems like it would make sense to me to have that implemented here instead of having to pass in the current time every time a new job is created in enforcer, since the current time will be almost always be used for new jobs and refunds.

ocflib/printing/quota.py
jvperrin: Yeah, I guess migrations would be a pain to do. I'm guessing we don't do a similar thing for the tables used in the [stats](https://github.com/ocf/ocflib/blob/master/ocflib/lab/stats.py) module because they were created using an init script in [ocfstats](https://github.com/ocf/labstats/blob/master/labstats/bin/init.py), but it seems like a pain to keep the schema here. Could it be moved into a separate file or something maybe to keep it a bit more separated (like in ocfstats), or do you think that would be a bad idea?

ocflib/printing/quota.py
chriskuehl: I agree with this, but unfortunately it's really hard to make namedtuples which accept default arguments without lots of boilerplate. I'd rather just ship it as-is if you're okay with that?

jvperrin: Yeah, that would be fine, we'll just have the current date duplicated in a couple places (utils, enforcer), and that will be fine. I would just ship it.

jvperrin:
Looks like it had trouble backing up last night, is the printed view incorrect somehow?

```
mysqldump: Couldn't execute 'SHOW FIELDS FROM `printed`': View 'ocfprinting.printed' references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them (1356)
----------------------------------------------------------------------------
rsnapshot encountered an error! The program was invoked with these options:
/usr/bin/rsnapshot -c /opt/share/backups/rsnapshot.conf sync
----------------------------------------------------------------------------
ERROR: backup_script /opt/share/backups/backup-mysql returned 2
WARNING: Rolling back "mysql/"
```

jvperrin:
Looks like the `printed_today` view is missing in the `ocfprinting` database right now, so any queries to the `printed` view is failing. Is there any reason why it's not present right now? I fixed it by adding back the view using [the schema](https://github.com/ocf/ocflib/pull/20/files#diff-2c3ec7f08522ddb82f2893546344cf99R190), but it looks like `printed` is still broken since all users and the `today` column are `NULL`.

ocflib/printing/quota.py
```diff
@@ -0,0 +1,236 @@
...
+            FROM jobs_semester AS jobs
+            LEFT OUTER JOIN refunds_semester AS refunds
+            ON jobs.user = refunds.user
+            GROUP BY jobs.user
+            ORDER BY user;
+
+        DROP VIEW IF EXISTS printed;
+        CREATE VIEW `printed` AS
+            SELECT
+                printed_today.user AS user,
```
jvperrin: I think this should be `printed_semester.user AS user,`, since the user might not have printed today, and so their name will be `NULL` if that is the case.

ocflib/printing/quota.py
```diff
@@ -0,0 +1,236 @@
...
+                printed_today.today AS today,
```
jvperrin: This could be changed to

    CASE WHEN printed_today.today IS NULL THEN 0 ELSE printed_today.today END AS today,

if we wanted it to have a default value of 0 instead of NULL when the user hasn't printed today, but that might be a bit too complicated or messy to be in the database itself.

chriskuehl:
@jvperrin thanks for the comments

i noticed some other issues as well, going to figure out a reasonable way to get tests for this before i try to merge (and will address your comments as well)

ocflib/printing/quota.py
chriskuehl: changed it to:
```sql
SELECT                                                
    printed_semester.user AS user,                    
    COALESCE(printed_today.today, 0) AS today,        
    COALESCE(printed_semester.semester, 0) AS semester
```

and added a test that it works

ocflib/printing/quota.py
chriskuehl: good catch, done (and test added to assert that it works)

chriskuehl:
@jvperrin the backup failure is because i was manually messing with the database trying to fix some faulty schema.

decided that was a bad idea and wrote some proper tests now. i'm pretty confident that if these tests pass, the schema is correct. feel free to suggest additional cases i'm not thinking of.

chriskuehl:
(and this should be ready to ship once i fix a jenkins-only test failure)

jvperrin:
@chriskuehl Ok, sounds good!

PR #21, opened 2016-03-06
[WIP] Fix refunds without jobs

chriskuehl:
Right now because of the way the join works, we don't include refunds for users who have never printed in our calculations.

Here's a (currently failing) test for that. If we can get it to pass we should be able to consider this fixed and merge this test plus the fix.


chriskuehl:
I'll try to fix this now.

chriskuehl:
Tests pass so this should be good to merge. I'll do this after lab closes since I need to modify the view schema (not dangerous, but could cause some errors).

PR #23, opened 2016-03-16
Fix unicode errors with document names

chriskuehl:


jvperrin:
Looks good to me! Nice test cases :smile:

chriskuehl:
added some cat emoji tests too

PR #25, opened 2016-05-26
Remove colorama dependency. Close #18

### LINKED ISSUE #18: Remove colorama dep
We don't really need it, we only use it for the color code constants. It does a lot of other stuff (e.g. wrapping stdout for Windows support) which we don't need.

jvperrin:
Still ugly and a lot of repeated code, but I don't see a better way to do it without modifying `locals()`, which flake8 complained about in the past. I guess a separate class could be made, but then all the existing imports would have to be changed.

### DIFF
--- a/ocflib/misc/shell.py
+++ b/ocflib/misc/shell.py
@@ -5,9 +5,6 @@ import subprocess
 import sys
 import tempfile
 
-from colorama import Back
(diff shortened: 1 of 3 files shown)


chriskuehl:
this is great, thanks!

PR #26, opened 2016-07-15
Better email templates, customize new account email for groups

chriskuehl:
We now don't tell groups that they have free printing, and we add a paragraph about virtual hosting.


jvperrin:
Looks good to me!

PR #28, opened 2016-08-25
Switch back to VeryFascistCheck

daradib:
Originally I checked passwords with VeryFascistCheck, which adds
some basic checks in Python to FascistCheck. Let's switch back,
at least until we switch to a pure Python implementation.

Currently we use the default wordlist, see `man update-cracklib`,
which is generated from /usr/share/dict/* in cron.daily.
Later we should probably use a real corpus of known passwords.

### DIFF
--- a/ocflib/account/validators.py
+++ b/ocflib/account/validators.py
@@ -41,7 +41,7 @@ def validate_password(username, password, strength_check=True):
             raise ValueError('Password is too similar to username')
 
         try:
-            cracklib.FascistCheck(password)
+            cracklib.VeryFascistCheck(password)
         except ValueError as e:
             raise ValueError('Password problem: {}'.format(e))
 


chriskuehl:
Why?

daradib:
1. Why not?
2. `VeryFascistCheck` just adds [two Python functions](https://github.com/cracklib/cracklib/blob/master/src/python/cracklib.py).
3. We originally used `VeryFascistCheck` in Django validation, see d099462.

I don't feel strongly either way though.

chriskuehl:
I'm -0.5 on this. I believe our existing password checks are probably already too stringent (or possibly the wrong kind of stringent).

Password security is a function of lots of things external to the password itself. At some point, people are either going to start re-using their strong passwords from other services for the OCF (bad), write it down in their phone's insecure notes app (bad), email it to themselves (bad), forget it (bad, though less bad), etc. And it leads to a really confusing experience at staff hours and for pubstaff who have to fend off hordes of people who forgot their password.

Is there evidence that the existing checks are insufficient? Keep in mind that from the passwords I was able to recently guess, all but 2 of them were forbidden by our current password checks (and we can supplement that with a list of common passwords at some point). I can think of only one case where we believe somebody's account was compromised due to a guessed password.

daradib:
I agree with you.

However, if you look at `VeryFascistCheck`, all it does is add a check for [palindromes](https://en.wikipedia.org/wiki/Palindrome). `pam_cracklib` does the same thing. The `simple` function doesn't apply to 8+ character passwords.

If you're still -0.5, I'll close.

chriskuehl:
I think my thoughts above aren't really applicable to this PR, just thoughts in general for when we revisit this.

daradib:
>  thoughts in general for when we revisit this

+1

PR #29, opened 2016-08-29
Add functions to modify LDAP attributes

matthew-mcallister:
In particular, add a celery task.

~~At the moment, it fails its most useful function in that it won't work properly with the mail attribute which requires a privileged bind to read. This will probably have to be solved by scraping the output of `ldapsearch`.~~

~~It also may be confusing for `create_ldap_entry_with_keytab` to also modify LDAP entries now.~~

### DIFF
--- a/ocflib/account/manage.py
+++ b/ocflib/account/manage.py
@@ -3,8 +3,11 @@ not account creation (since it's too large)."""
 import pexpect
 
 import ocflib.account.search as search
+import ocflib.account.utils as utils
 import ocflib.account.validators as validators
 import ocflib.constants as constants
+import ocflib.infra.ldap as ldap_ocf
+import ocflib.misc as misc
 import ocflib.misc.mail as mail
 import ocflib.misc.shell as shell
 
@@ -79,6 +82,26 @@ def change_password_with_keytab(username, password, keytab, principal, comment=N
         raise ValueError('kadmin Error: {}'.format(output))
 
     _notify_password_change(username, comment=comment)
+
+
+def modify_ldap_attributes(username, attributes, keytab, principal):
+    """Adds or modifies arbitrary attributes of a user's LDAP record subject to
+    minor validation beyond the LDAP schema.
+
+    At the moment, the only attribute that benefits from extra validation is
+    the 'loginShell' attribute.
+    """
+
+    for value in attributes.get('loginShell', ()):
+        if not misc.validators.valid_login_shell(value):
+            raise ValueError("Invalid login shell '{}'".format(value))
+
+    ldap_ocf.modify_ldap_entry_with_keytab(
+        utils.dn_for_username(username),
+        attributes,
+        keytab,
+        principal,
+    )
 
 
 def _notify_password_change(username, comment=None):
--- a/ocflib/account/utils.py
+++ b/ocflib/account/utils.py
@@ -146,3 +146,10 @@ def list_staff(group='ocfstaff'):
     :param group: UNIX group to use to determine if someone is a staff member.
     """
     return grp.getgrnam(group).gr_mem
+
+
+def dn_for_username(username):
+    return 'uid={user},{base_people}'.format(
+        user=username,
+        base_people=constants.OCF_LDAP_PEOPLE,
+    )
--- a/ocflib/infra/ldap.py
+++ b/ocflib/infra/ldap.py
@@ -52,43 +52,33 @@ def ldap_ucb():
     return ldap_connection(constants.UCB_LDAP)
 
 
-def create_ldap_entry_with_keytab(
-    dn,
-    attributes,
-    keytab,
-    admin_principal,
-):
-    """Creates an LDAP entry by shelling out to ldapadd.
-
-    :param dn: distinguished name of the new entry
-    :param attributes: dict mapping attribute name to list of values
-    :param keytab: path to the admin keytab
-    :param admin_principal: admin principal to use with the keytab
-    """
+def _format_attr(key, values):
     # LDAP attributes can have multiple values, but commonly we don't consider
     # that. So, let's sanity check the types we've received.
-    for v in attributes.values():
-        assert type(v) in (list, tuple), 'Value must be list or tuple.'
+    assert type(values) in (list, tuple), 'Value must be list or tuple.'
 
-    def format_attr(key, values):
-        # might be possible to have non-ASCII letters in keys, but don't think
-        # it will happen to us. we can fix this if it ever does.
-        assert all(c in string.ascii_letters for c in key), 'key is not ASCII letters'
+    # might be possible to have non-ASCII letters in keys, but don't think
+    # it will happen to us. we can fix this if it ever does.
+    assert all(c in string.ascii_letters for c in key), 'key is not ASCII letters'
 
-        # rather than try to carefully escape values, we just base64 encode
-        return (
-            '{key}:: {value}'.format(
-                key=key,
-                value=b64encode(value.encode('utf8')).decode('ascii'),
-            ) for value in values
-        )
+    lines = [
+        '{key}:: {value}'.format(
+            key=key,
+            value=b64encode(value.encode('utf8')).decode('ascii'),
+        ) for value in values
+    ]
 
-    lines = list(chain(
-        format_attr('dn', [dn]),
-        *(format_attr(key, values) for key, values in attributes.items())
-    ))
+    return lines
 
-    cmd = 'kinit -t {keytab} {principal} ldapadd'.format(
+
+def _write_ldif(lines, dn, keytab, admin_principal):
+    """Issue an update to LDAP via ldapmodify in the form of lines of an LDIF
+    file.
+
+    :param lines: ldif file as a sequence of lines
+    """
+
+    cmd = 'kinit -t {keytab} {principal} ldapmodify'.format(
         keytab=keytab,
         principal=admin_principal,
     )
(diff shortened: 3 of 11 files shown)


ocflib/account/manage.py
```diff
@@ -81,6 +84,54 @@ def change_password_with_keytab(username, password, keytab, principal, comment=N
...
+    """Adds or modifies arbitrary attributes of a user's LDAP record subject to
+    basic validation.
+
+    Attribute values which are not a list or tuple will be wrapped in one.
+    Validation is applied for, e.g., the mail and loginShell fields, but not
+    for all fields.
+    """
+    for k, v in attributes.items():
+        if not type(v) in (list, tuple):
+            attributes[k] = (v,)
```
chriskuehl: I think it would be much better to raise on errors rather than try to correct them. The caller should use the function properly. This type of pattern leads to many extra code paths that make it difficult to use and test the code fully, and I frequently see it result in unexpected bugs later on.

This is what we do in another function:

```python
# LDAP attributes can have multiple values, but commonly we don't consider
# that. So, let's sanity check the types we've received.                  
for v in attributes.values():                                             
    assert type(v) in (list, tuple), 'Value must be list or tuple.'       
```

ocflib/account/manage.py
```diff
@@ -81,6 +84,54 @@ def change_password_with_keytab(username, password, keytab, principal, comment=N
...
+
+    SINGLE_VALUED = {'mail', 'loginShell', 'calnetUid', 'callinkOid'}
+
+    for key in SINGLE_VALUED:
+        if len(attributes.get(key, ())) > 1:
+            raise ValueError("Attribute '{}' is single-valued".format(key))
```
chriskuehl: I don't like the special cases here for certain attributes. Better to let the LDAP server be the source of truth for this. Attributes can be declared single-value in the LDAP schema if that is our intention.

ocflib/account/manage.py
```diff
@@ -81,6 +84,54 @@ def change_password_with_keytab(username, password, keytab, principal, comment=N
...
+        return True
+
+    EXTRA_VALIDATORS = {
+        ('mail', misc.validators.valid_email,
+            "Invalid email address supplied to 'mail' attribute"),
+        ('loginShell', misc.validators.valid_login_shell,
+            "Invalid login shell path supplied to 'loginShell' attribute"),
+        ('calnetUid', is_int, "Non-number supplied to 'calnetUid' attribute"),
+        ('callinkOid', is_int, "Non-number supplied to 'callinkOid' attribute"),
+    }
```
chriskuehl: I think this function is trying to do too much. Compare this to `create_ldap_entry_with_keytab` which doesn't try to understand the account model at all and just does what you ask it to.

Almost all of this validation is provided by the LDAP server anyway. I think the only exception is `loginShell`, and `ocfweb` should probably just handle that (this is what `chsh` does currently).

ocflib/account/utils.py
```diff
@@ -146,3 +146,10 @@ def list_staff(group='ocfstaff'):
     :param group: UNIX group to use to determine if someone is a staff member.
     """
     return grp.getgrnam(group).gr_mem
+
+
+def dn_for_username(username):
+    return 'uid={user},{base_people}'.format(
+        user=username,
+        base_people=constants.OCF_LDAP_PEOPLE,
+    )
```
chriskuehl: nice

ocflib/infra/ldap.py
```diff
@@ -57,40 +57,72 @@ def create_ldap_entry_with_keytab(
     attributes,
     keytab,
     admin_principal,
+    modify=False,
 ):
-    """Creates an LDAP entry by shelling out to ldapadd.
+    """Creates or modifies an LDAP entry by shelling out to ldapadd/ldapmodify.
```
chriskuehl: I would much prefer a second function for this. There's almost no shared code, and what is can be split out into helper functions. These are two distinct actions and it makes a lot of sense to have two distinct functions.

ocflib/infra/ldap.py
```diff
@@ -57,40 +57,72 @@ def create_ldap_entry_with_keytab(
...
+                    '-',
+                ) if key in existing_attrs else (),
+                *(
+                    (
+                        'add: {}'.format(key),
+                        encode(key, value),
+                        '-',
+                    ) for value in values
+                )
+            )
```
chriskuehl: If you use `replace` instead I think you can eliminate almost all of this (plus the checking for the entry beforehand, which eliminates another race condition too):

http://serverfault.com/a/471168

ocflib/account/manage.py
matthew-mcallister: I didn't know that the LDAP schema specifies all this information. Now that I realize this, I agree that all the validation is pointless. This was all originally thrown in just because I had the idea.

ocflib/misc/validators.py
```diff
@@ -39,3 +39,23 @@ def valid_email(email):
...
+        '/bin/rbash',
+        '/usr/bin/screen',
+        '/usr/bin/tmux',
+        '/bin/zsh',
+        '/usr/bin/zsh',
+        '/bin/tcsh',
+        '/usr/bin/tcsh',
+    }
+
+    return shell in VALID_LOGIN_SHELLS
```
chriskuehl: Probably valuable to expose this list at the module level? I can imagine ocfweb wanting to use it in a drop-down.

tests/account/manage_test.py
```diff
@@ -87,6 +96,40 @@ def test_kadmin_failure(self, mock_spawn, mock_notify_password_change):
...
+        with pytest.raises(ValueError):
+            modify_ldap_attributes(
+                'ggroup',
+                attrs,
+                '/some/keytab',
+                'create/admin',
+            )
+        assert not mock_create_ldap_entry_with_keytab.called
+
+
```
chriskuehl: nice tests

ocflib/account/manage.py
matthew-mcallister: That was only added because I thought it would make the interface more convenient. If it seems like a bad idea to you, I'm sure you're right.

tests/account/utils_test.py
```diff
@@ -233,3 +234,8 @@ def test_list_staff():
     assert 'ckuehl' in staff
     assert 'bpreview' not in staff
     assert 5 <= len(staff) <= 50
+
+
+def test_dn_for_username():
+    assert (dn_for_username('kpengboy') ==
+            'uid=kpengboy,ou=People,dc=OCF,dc=Berkeley,dc=EDU')
```
chriskuehl: nice test

ocflib/misc/validators.py
matthew-mcallister: Good idea.

ocflib/infra/ldap.py
matthew-mcallister: Well shoot, that would be a major improvement. I thought replace was basically the same as delete followed by add.

ocflib/infra/ldap.py
matthew-mcallister: Yeah... this was just sloppiness. It's easier to start by either copy-pasting or modifying an existing function, and, before I started, I thought it wouldn't be too much more involved than replacing `ldapadd` with `ldapmodify` and maybe switching a command line flag.

matthew-mcallister:
Comments addressed.

ocflib/account/manage.py
```diff
@@ -81,6 +84,28 @@ def change_password_with_keytab(username, password, keytab, principal, comment=N
...
+    minor validation beyond the LDAP schema.
+
+    At the moment, the only attribute that benefits from extra validation is
+    the 'loginShell' attribute.
+    """
+
+    if 'loginShell' in attributes:
+        for value in attributes['loginShell']:
+            if not misc.validators.valid_login_shell(value):
+                raise ValueError("Invalid login shell '{}'".format(value))
```
chriskuehl: could write as:

```python
        for value in attributes.get('loginShell', []):
            if not misc.validators.valid_login_shell(value):
                raise ValueError("Invalid login shell '{}'".format(value))
```

and eliminate the `if`. up to you.

ocflib/account/manage.py
```diff
@@ -81,6 +84,28 @@ def change_password_with_keytab(username, password, keytab, principal, comment=N
...
+
+    dn = utils.dn_for_username(username)
```
chriskuehl: nit: probably don't need this variable

ocflib/infra/ldap.py
```diff
@@ -52,43 +52,40 @@ def ldap_ucb():
...
+    ]
+
+    if action:
+        return chain(
+            ('{action}: {key}'.format(action=action, key=key),),
+            lines,
+            ('-',),
         )
+    else:
+        return lines
```
chriskuehl: feels to me like this action logic should probably just be put in the modify function since it's pretty specific to ldapmodify's input format. don't feel too strongly about it though.

ocflib/misc/validators.py
```diff
@@ -39,3 +39,27 @@ def valid_email(email):
...
+    '/usr/bin/tmux',
+    '/bin/zsh',
+    '/bin/tcsh',
+}
+
+# Separate the duplicates so we can list the above to users
+VALID_LOGIN_SHELLS_ALTPATHS = {
+    '/usr/bin/zsh',
+    '/usr/bin/tcsh',
+}
```
chriskuehl: can we make these `frozenset` objects so it's harder for people to mess with them?

```python
VALID_LOGIN_SHELLS = frozenset((
    'shell1',
    'shell2',
    ...,
))
```

ocflib/misc/validators.py
```diff
@@ -39,3 +39,27 @@ def valid_email(email):
...
+
+
+def valid_login_shell(shell):
+    """Test that a file path is an actual valid login shell on tsunami."""
+    return shell in VALID_LOGIN_SHELLS or shell in VALID_LOGIN_SHELLS_ALTPATHS
```
chriskuehl: you can write:

```python
return shell in (VALID_LOGIN_SHELLS | VALID_LOGIN_SHELLS_ALTHPATHS)
```

if you want

chriskuehl:
lgtm with a couple comments, no blockers

matthew-mcallister:
Thanks for the extra comments. I took all the minor suggestions because I'm sure your python is more idiomatic than mine.

PR #32, opened 2016-10-16
account.create: cache min uid in redis

matthew-mcallister:
This makes a lot more sense to me than caching in the module itself, though passing parameters through multiple function calls is slightly less than elegant. I'm not sure how to test this though.

See https://github.com/ocf/ocflib/pull/31 for @daradib's PR which led me to create this one.

### DIFF
--- a/ocflib/account/creation.py
+++ b/ocflib/account/creation.py
@@ -26,8 +26,42 @@ from ocflib.misc.validators import valid_email
 from ocflib.printing.quota import SEMESTERLY_QUOTA
 
 
-def create_account(request, creds, report_status):
-    """Create an account as idempotently as possible."""  # TODO: docstring
+_KNOWN_UID = 43000
+
+
+def _get_first_available_uid(known_uid=_KNOWN_UID):
+    """Return the first available UID number.
+
+    Searches our entire People ou in order to find it. It seems like there
+    should be a better way to do this, but quick searches don't show any.
+
+    We hard-code a value we know has already been reached and only select
+    entries greater than that for performance. This value can then be cached
+    and passed back in to make subsequent calls faster.
+    """
+    with ldap_ocf() as c:
+        c.search(
+            constants.OCF_LDAP_PEOPLE,
+            '(uidNumber>={KNOWN_MIN})'.format(KNOWN_MIN=known_uid),
+            attributes=['uidNumber'],
+        )
+        uids = [int(entry['attributes']['uidNumber'][0]) for entry in c.response]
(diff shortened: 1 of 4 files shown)


ocflib/account/creation.py
```diff
@@ -43,7 +49,10 @@ def create_account(request, creds, report_status):
 
     # TODO: check if LDAP entry already exists; skip this if so
     with report_status('Finding', 'Found', 'first available UID'):
-        new_uid = _get_first_available_uid()
+        try:
+            new_uid = _get_first_available_uid(kwargs['known_uid'])
+        except KeyError:
+            new_uid = _get_first_available_uid()
```
chriskuehl: i would suggest to just add `known_uid=_KNOWN_UID` to the function signature, and then you can just always use it in the function call here (and don't need to use `**kwargs`)

tests/account/submission_test.py
```diff
@@ -266,6 +266,7 @@ def test_create_no_issues(
                 fake_credentials,
                 mock.ANY,
             )
+            mock_redis_locking.return_value.set.assert_called_once_with('known_uid', mock.ANY)
```
chriskuehl: you can write `mock_redis_locking().set` instead of `mock_redis_locking.return_value.set` if you want

ocflib/account/creation.py
```diff
@@ -26,8 +26,14 @@
 from ocflib.printing.quota import SEMESTERLY_QUOTA
 
 
-def create_account(request, creds, report_status):
-    """Create an account as idempotently as possible."""  # TODO: docstring
+def create_account(request, creds, report_status, **kwargs):
+    """Create an account as idempotently as possible.
+
+    :kwarg known_uid: where to start searching for unused UIDs (see
```
matthew-mcallister: How are you actually supposed to document kwargs?

ocflib/account/creation.py
matthew-mcallister: Yeah, I don't think using kwargs really adds anything here.

ocflib/account/creation.py
chriskuehl: i usually just use `arg` since kwargs in python are barely different than regular args (you can use `argname=value` for positional arguments too, you can fill in kwargs without specifying the keyword name, etc.), but there are surely more formal ways to do it. this seems fine to me.

you can get really fancy with the various docstring-to-documentation generators, but afaik most effectively require writing rst which is enough to make you want to stop being a programmer

PR #34, opened 2016-10-19
Make account creation more idempotent

matthew-mcallister:
I'm sure this will accrue many nits.

Still need to add manual tests for `get_kerberos_principal_with_keytab`.

### DIFF
--- a/ocflib/account/creation.py
+++ b/ocflib/account/creation.py
@@ -1,7 +1,7 @@
 import difflib
 import itertools
 import math
-import os.path
+import os
 import re
 import subprocess
 from collections import namedtuple
@@ -18,6 +18,7 @@ import ocflib.account.utils as utils
 import ocflib.account.validators as validators
 import ocflib.constants as constants
 from ocflib.infra.kerberos import create_kerberos_principal_with_keytab
+from ocflib.infra.kerberos import get_kerberos_principal_with_keytab
 from ocflib.infra.ldap import create_ldap_entry_with_keytab
 from ocflib.infra.ldap import ldap_ocf
 from ocflib.misc.mail import jinja_mail_env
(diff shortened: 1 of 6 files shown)


ocflib/account/submission.py
```diff
@@ -228,18 +228,18 @@ def create_account(request):
             # status reporting
             status = []
 
-            def _report_status(line):
+            def report_status(line):
                 """Update task status by adding the given line."""
                 status.append(line)
                 create_account.update_state(meta={'status': status})
```
chriskuehl: should we just make this silly thing a combination context manager / function?
```python
class report_status:

    def __init__(self, *args):
        if len(args) == 1:
            self(*args)
        else:
            self.start, self.stop, self.task = args

    def __call__(self, msg):
        print(msg)

    def __enter__(self, *args):
        self(self.start + ' ' + self.task)

    def __exit__(self, *args):
        self(self.stop + ' ' + self.task)


report_status('hello world')

with report_status('started', 'stopped', 'thing'):
    time.sleep(1)
```

kinda magical, but we can avoid re-defining it over-and-over

ocflib/account/creation.py
```diff
@@ -117,37 +140,42 @@ def create_account(request, creds, report_status, known_uid=_KNOWN_UID):
...
+        ])
+
+    public_html_path = utils.public_html_path(user)
+
+    if not os.path.exists(public_html_path):
+        # symlink it from ~user/public_html
+        subprocess.check_call([
+            'sudo', '-u', user,
+            'ln', '-fs', '--', path, public_html_path,
+        ])
```
chriskuehl: this is unnecessary, `-f` to `ln` makes it idempotent

ocflib/account/creation.py
```diff
@@ -117,37 +140,42 @@ def create_account(request, creds, report_status, known_uid=_KNOWN_UID):
...
+    # other files named public_html, particularly when unsorrying accounts
     path = utils.web_dir(user)
 
-    # create web directory
-    subprocess.check_call([
-        'sudo', 'install',
-        '-d', '--mode=0755', '--group=ocf', '--owner=' + user,
-        '--',
-        path,
-    ])
```
chriskuehl: pretty sure install is idempotent already

ocflib/account/creation.py
```diff
@@ -117,37 +140,42 @@ def create_account(request, creds, report_status, known_uid=_KNOWN_UID):
...
+    """Create home directory for user with appropriate permissions."""
     home = utils.home_dir(user)
     subprocess.check_call(
         ['sudo', 'install', '-d', '--mode=0700', '--group=ocf',
             '--owner=' + user, home])
 
 
-def create_web_dir(user):
-    """Create web directory for user with appropriate permissions.
+def rectify_web_dir(user):
```
chriskuehl: really useful function

what about calling it just `ensure_web_dir` though?

ocflib/account/creation.py
```diff
@@ -117,37 +140,42 @@ def create_account(request, creds, report_status, known_uid=_KNOWN_UID):
 
 
 def create_home_dir(user):
-    """Create home directory for user. Makes a directory with appropriate
-    permissions, then copies in OCF's skeleton dotfiles.
-    """
+    """Create home directory for user with appropriate permissions."""
```
chriskuehl: nice

ocflib/infra/kerberos.py
```diff
@@ -54,3 +53,34 @@ def create_kerberos_principal_with_keytab(
...
+    child = pexpect.spawn(cmd, timeout=10)
+    child.expect(pexpect.EOF)
+
+    output = child.before.decode('utf8')
+    if 'Principal does not exist' in output:
+        return None
+    elif 'kadmin' in output:
+        raise ValueError('kadmin Error: {}'.format(output))
+
+    return True
```
chriskuehl: can we check the exit code of this instead? that would probably be more reliable than reading the stdout

tests/account/creation_test.py
```diff
@@ -493,13 +500,14 @@ def test_create(
         fake_new_account_request,
         fake_credentials
     ):
-        @contextmanager
-        def report_status(start, end, line):
-            yield
+        def report_status(line):
+            pass
```
chriskuehl: might as well just use `mock.Mock` if you don't need the context manager

ocflib/infra/kerberos.py
matthew-mcallister: You know, I actually ran kadmin to check that it gave a nonzero exit code when the principal wasn't found, then promptly forgot and wrote that junk in.

ocflib/account/creation.py
matthew-mcallister: Does `-f` only remove the target if it's a symlink or does it delete files? Documentation isn't clear.

ocflib/account/submission.py
matthew-mcallister: Yeah, I was originally going to make it act as a context manager when passed three arguments and a single line reporter when passed one, but I didn't feel like looking up how context managers actually work to do that.

ocflib/account/creation.py
chriskuehl: it will remove files but not directories

matthew-mcallister: Either way, the function would ideally be augmented to rename existing files to make room for the symlink. I can imagine at least one user has a folder named `public_html` in their home directory, and it would be great if `makehttp` would fix that up..

ocflib/infra/kerberos.py
matthew-mcallister: Don't we want to raise an error if kdamin returns some error message besides "principal not found"? If so, checking the exit code doesn't really help.

matthew-mcallister: Oh I see, you mean to catch all other errors.

ocflib/infra/kerberos.py
```diff
@@ -50,7 +49,37 @@ def create_kerberos_principal_with_keytab(
...
+             None if the principal does not exist
+    """
+    cmd = ('{kadmin} -K {keytab} -p {admin} get {principal}').format(
+        kadmin=escape_arg(KADMIN_PATH),
+        keytab=escape_arg(keytab),
+        admin=escape_arg(admin_principal),
+        principal=escape_arg(principal),
+    )
+
+    output, status = pexpect.run(cmd, timeout=10, withexitstatus=True)
```
matthew-mcallister: Should I prefer `subprocess.check_output` to `pexpect.run`?

matthew-mcallister:
Should have addressed all comments at this point.

PR #41, opened 2016-11-18
Bugfix and tests for ucb.directory

matthew-mcallister:
This corrects `signatories_for_group` to get names using `name_by_calnet_uid`, fixing a bug with the new version of ldap3, and adds tests for the functions in `ucb.directory`.

I would like to test `ucb.groups` as well, but those will likely be flakier if we want to actually query CalLink.

### DIFF
--- a/ocflib/ucb/directory.py
+++ b/ocflib/ucb/directory.py
@@ -34,7 +34,7 @@ def name_by_calnet_uid(uid):
         if given_name and sn:
             return '{} {}'.format(given_name, sn)
     else:
-        return names.get('displayName', None)
+        return names.get('displayName')
 
 
 def calnet_uids_by_name(name):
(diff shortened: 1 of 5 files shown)


ocflib/ucb/directory.py
```diff
@@ -33,11 +33,8 @@ def get_longest_string(strs):
 
         if given_name and sn:
             return '{} {}'.format(given_name, sn)
-    elif 'displayName' in names:
-        display_name = get_longest_string(names['displayName'])
-
-        if display_name:
-            return display_name
+    else:
+        return names.get('displayName', None)
```
chriskuehl: `None` is the default

tests/conftest.py
```diff
@@ -10,6 +10,15 @@
 import pytest
 
 
+# To test functions that query UCB LDAP for people, we simply pick
+# someone who has a few years before graduating and use their name and
+# CalNet UID. Alumni eventually get kicked out of the university's
+# "People" OU, so these constants have to be updated every few years
+# after the aforementioned party has graduated.
+TEST_PERSON_CALNET_UID = 1101587
+TEST_PERSON_NAME = 'Jason Perrin'
```
chriskuehl: jvperrin can never graduate now

PR #45, opened 2017-01-13
Add Jenkinsfile

jvperrin:
This adds a Jenkinsfile for this repo, pretty much the same as the [one in pam_trimspaces](https://github.com/ocf/pam_trimspaces/blob/master/Jenkinsfile).


jvperrin:
This also need to have triggers for create, ocfweb, etc, but those also still need to be added to other Jenkinsfiles in other repos, so I figured adding them all at once would be best there.

Jenkinsfile
```diff
@@ -0,0 +1,71 @@
...
+            sh "make package_${dist}"
+            sh "mv dist dist_${dist}"
+            archiveArtifacts artifacts: "dist_${dist}/*"
+        }
+
+        stash 'src'
+    }
+
+    if (env.BRANCH_NAME == 'master') {
+        stage name: "upload-${dist}"
```
chriskuehl: i think the new way to do this is like

```groovy
stage("upload-${dist}") {
    build job: 'upload-changes', parameters: [
         [$class: 'StringParameterValue', name: 'path_to_changes', value: "dist_${dist}/*.changes"],
         [$class: 'StringParameterValue', name: 'dist', value: dist],
         [$class: 'StringParameterValue', name: 'job', value: env.JOB_NAME.replace('/', '/job/')],
         [$class: 'StringParameterValue', name: 'job_build_number', value: env.BUILD_NUMBER],
    ]
}
```

but this is ok for now

Jenkinsfile
```diff
@@ -0,0 +1,71 @@
...
+        dir('src') {
+            checkout scm
+        }
+    }
+
+    withCredentials([[
+        $class: 'StringBinding',
+        credentialsId: 'coveralls_token',
+        variable: 'COVERALLS_REPO_TOKEN'
+    ]]) {
```
chriskuehl: neat

Jenkinsfile
jvperrin: Alright, sounds good, this would be good to do across all the Jenkinsfiles if that's the new way of doing things.

PR #47, opened 2017-01-24
staff_in_lab_count

tahabi:
with reference to https://github.com/ocf/labstats/pull/2

### DIFF
--- a/ocflib/lab/stats.py
+++ b/ocflib/lab/stats.py
@@ -68,9 +68,8 @@ def staff_in_lab():
 
 def staff_in_lab_count():
     """Return the count of unique staff in the lab"""
-    # thank SO: http://stackoverflow.com/questions/10024646/how-to-get-unique-list-using-a-key-word-python
-    seen = set()
-    return len([seen.add(staff.user) or staff for staff in staff_in_lab() if staff.user not in seen])
+    # thank based ckuehl https://github.com/ocf/ocflib/pull/47#discussion_r97914296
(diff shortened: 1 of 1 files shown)


ocflib/lab/stats.py
```diff
@@ -66,6 +66,13 @@ def staff_in_lab():
         return [Session.from_row(r) for r in c]
 
 
+def staff_in_lab_count():
+    """Return the count of unique staff in the lab"""
+    # thank SO: http://stackoverflow.com/questions/10024646/how-to-get-unique-list-using-a-key-word-python
+    seen = set()
+    return len([seen.add(staff.user) or staff for staff in staff_in_lab() if staff.user not in seen])
```
chriskuehl: this is insane, please don't commit this

can't you just write `return len({session.user for session in staff_in_lab()})`?

tahabi: lol that is a much better solution

ocflib/lab/stats.py
```diff
@@ -66,6 +66,12 @@ def staff_in_lab():
         return [Session.from_row(r) for r in c]
 
 
+def staff_in_lab_count():
+    """Return the count of unique staff in the lab"""
+    # thank based ckuehl https://github.com/ocf/ocflib/pull/47#discussion_r97914296
```
kpengboy: I would say the comment would be superfluous in the long run...

tahabi: one must atone for one's sins

PR #53, opened 2017-02-24
creating a mysql view for storing public jobs data

shex1627:
I am trying to create a histogram of the frequency of each print jobs' page. I made an sql query(on the jobs database) so that it only contains

1.day
2.pages(from 1 to 20, right now i dont know how to adjust it to maximum quota)
3.count(number of jobs print that amout of pages)

### DIFF
--- a/ocflib/printing/ocfprinting.sql
+++ b/ocflib/printing/ocfprinting.sql
@@ -125,4 +125,14 @@ CREATE VIEW `printed` AS
     ON printed_today.user = printed_semester.user
     ORDER BY user;
 
+DROP VIEW IF EXISTS public_jobs;
+CREATE VIEW `public_jobs` AS
+    SELECT
+        DATE_FORMAT(`time`, '%Y-%m') AS `day`, `pages`, COUNT(`pages`) AS `count`
+    FROM `jobs`
+    WHERE `pages` <= 20 AND `pages` > 0
+    GROUP BY `day`, `pages`
+    ORDER BY `day` DESC
+
 GRANT SELECT ON `ocfprinting`.`printed` TO 'anonymous'@'%';
+GRANT SELECT on `ocfprinting`.`public_jobs` TO 'anonymous'@'%';
--- a/ocflib/printing/ocfprinting.sql
+++ b/ocflib/printing/ocfprinting.sql
@@ -128,9 +128,9 @@ CREATE VIEW `printed` AS
 DROP VIEW IF EXISTS public_jobs;
 CREATE VIEW `public_jobs` AS
     SELECT
-        DATE_FORMAT(`time`, '%Y-%m') AS `day`, `pages`, COUNT(`pages`) AS `count`
+        DATE_FORMAT(`time`, '%Y-%m-%d') AS `day`, `pages`, COUNT(`pages`) AS `count`
     FROM `jobs`
-    WHERE `pages` <= 20 AND `pages` > 0
+    WHERE `pages` > 0
     GROUP BY `day`, `pages`
     ORDER BY `day` DESC
 


jvperrin:
If you want to adjust it to the maximum quota, you can have the view contain all the jobs, and then only select the ones from the view that are within the quota, where the quotas are defined [here](https://github.com/ocf/ocflib/blob/cb8d08c3e54889936e4c8f053492891e54350ff8/ocflib/printing/quota.py#L11). Might be easier/cleaner than limiting it to only jobs less than 20 pages anyway.

Also are there really jobs with negative numbers of pages printed? O.o

jvperrin:
Also add your OCF email to your GitHub account so that you get credit for the commit! :D

shex1627:
Regarding changing max quota:
We store the max quota in a python file but I am editing an sql file though. 

Also, we have very small amount of page  0 jobs that is why I added "'pages' > 0"

chriskuehl:
lgtm but i wonder why you need the `<= 20` at all? does it matter if we don't specify a max?

jvperrin:
Oh I see, so not negative but zero, that makes sense.

That's what I'm saying, you make the view so that it contains all the jobs, not just the ones 20 pages or under, and then when you query the view, then you can specify whatever max you want.

shex1627:
Regarding <= 20:
There are very few print jobs that are 40+,50+ or 80+(probably staff) and I am trying to hide that.

NickImpicciche:
+1 for hiding

On Feb 24, 2017 3:53 PM, "shicheng" <notifications@github.com> wrote:

Regarding <= 20:
There are very few print jobs that are 40+,50+ or 80+(probably staff) and I
am trying to hide that.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/ocf/ocflib/pull/53#issuecomment-282437094>, or mute the
thread

jvperrin:
Ok, sure, but if the printing quota changes later on, that might not be the case, and the number will have to be adjusted when the ocflib printing quota constants can be used instead to dynamically select only the jobs that could be printed within the quota.

chriskuehl:
why hide? it doesn't seem worth it to me, nobody can see what the jobs are, and in practice nobody's gonna query the database anyway

NickImpicciche:
Using the constants makes sense, I guess in practice it doesn't really
matter, but I'm not opposed to hiding jobs not in quota either.

chriskuehl:
i'd agree to using the constants in the code if we could access them from sql but we can't :(

jvperrin:
The constants can be used when making the queries though, so I think they should be used there instead.

chriskuehl:
agreed, it just doesn't achieve the goal of hiding sketchy jobs with lots of pages from anybody who queries the database

jvperrin:
Maybe people shouldn't have printed huge jobs if they didn't want them to be in the database ;)

shex1627:
So should I merge the pull request now or take out the ">= 20"

ocflib/printing/ocfprinting.sql
```diff
@@ -125,4 +125,14 @@ CREATE VIEW `printed` AS
     ON printed_today.user = printed_semester.user
     ORDER BY user;
 
+DROP VIEW IF EXISTS public_jobs;
+CREATE VIEW `public_jobs` AS
+    SELECT
+        DATE_FORMAT(`time`, '%Y-%m') AS `day`, `pages`, COUNT(`pages`) AS `count`
```
tahabi: shouldn't this be 'AS `month`'?

tahabi:
Perhaps something like:

"SELECT DATE_FORMAT(`time`, '%Y-%m') AS `month`, `pages`, COUNT(`pages`) AS `n`, SUM(`pages`) AS `total` FROM `jobs` WHERE `pages` > 0 GROUP BY `month`, `pages` ORDER BY `month` DESC"

shex1627:
Oh I messed up. I am supposed to format the 'time' to day instead of month. So if I want to make a change, do I push against and start a new pull request?

tahabi:
push commits to the jobsview branch on your fork of ocflib

shex1627:
1. I correct the time format to `day` instead of `month` originally. 
2. I remove the " <= 20" because even so there are still jobs during the weekdays that is more than 10 pages. (dealing with those jobs is just a pain and has little practical value)
So I think it is ready to send my merge pull request，choosing the rebase and merge option)?

chriskuehl:
it has been applied to prod

PR #54, opened 2017-02-26
Parallelize jessie and stretch builds

jvperrin:
Hopefully this works, I've seen things that say that there can't be multiple stages within a parallel block, so we'll have to see if that actually works or not.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -43,27 +43,31 @@ def branches = [:]
 for (def i = 0; i < dists.size(); i++) {
     def dist = dists[i]
     branches["build-and-deploy-${dist}"] = {
-        node('slave') {
-            step([$class: 'WsCleanup'])
-            unstash 'src'
+        stage("build-${dist}") {
+            node('slave') {
+                step([$class: 'WsCleanup'])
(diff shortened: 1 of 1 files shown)


jvperrin:
ERROR: The ‘stage’ step must not be used inside a ‘parallel’ block.

Yup, well, guess it didn't work...

chriskuehl:
It doesn't work with the "old-style" stage lines (where the contents of the stage aren't inside a block), but works if you use `stage` blocks. See e.g. ocfweb: https://github.com/ocf/ocfweb/blob/1655af5d6a73323e895eab44980e559e1f60fcf9/Jenkinsfile#L43-L66

jvperrin:
Ah, I see, I looked at that before but failed to see the `stage` blocks within that, nice

jvperrin:
Hmm, strange that it doesn't show either of the parallel steps: https://jenkins.ocf.berkeley.edu/job/ocflib/view/change-requests/job/PR-54/

jvperrin:
Well, that seems to work ok, they do run in parallel, although the interface is still confusing (as it was before).

PR #56, opened 2017-03-04
finished moving constants to its respective files

MrMinos:
from … RT#5941

as noted in the ticket #5941, not sure about moving whole RESERVED_USERNAMES to ocflib/account/validators.py file, because that list is about 260 lines. Waiting on feedback about what to do. But all the other files in ocflib repository now should be using its own file's constants.


tahabi:
You can make the change, Chris mentioned it's only used in that one file.

Also you need to fix the build: there's a syntax error that's causing tests to fail.

MrMinos:
comma being at ocflib/account/utils.py is correct. I just put newline in the wrong place. Fixed it : )

ocflib/account/search.py
```diff
@@ -1,15 +1,17 @@
 """Methods for searching and selecting users."""
 import ldap3
 
-import ocflib.constants as constants
 import ocflib.infra.ldap as ldap
 
+OCF_LDAP_PEOPLE = 'ou=People,dc=OCF,dc=Berkeley,dc=EDU'
+UCB_LDAP_PEOPLE = 'ou=People,dc=Berkeley,dc=EDU'
```
MrMinos: sounds good. since I need to leave ocf now I'll work on merging some of the duplicated constants and push it all together later.

MrMinos: Moved most if not all LDAP constants to ocflib/infra/ldap.py and when needed in other files you would import the constants from said mentioned ldap.py file.

tahabi:
You don't need to do that, just log into Jenkins and re-trigger the job.

MrMinos:
@tahabi I was looking for such functionality but I could not find it on my dashboard when I log into Jenkins
screen shot 2017-03-04 at 15 09 13

tahabi:
screenshot_20170304-151422

You have to log in with your OCF account credentials

MrMinos:
Yea I did but I do not see it. I am guessing permission is not granted on my account.

Thanks for the feedback.

tahabi:
Make sure you're on the page for your PR. No extra permissions should be
necessary.

MrMinos:
Still don't see it..
I am at this address:
https://jenkins.ocf.berkeley.edu/job/ocflib/job/PR-56/

screen shot 2017-03-04 at 15 35 42

And at this address:
https://jenkins.ocf.berkeley.edu/job/ocflib/job/PR-56/4/
screen shot 2017-03-04 at 15 34 57

MrMinos:
@tahabi please review my changes. Thank you!

jvperrin:
@tahabi Only `ocfroot` users can trigger builds I believe, since having access to manage Jenkins allows for deploying software pretty much anywhere in the OCF, which could be misused.

chriskuehl:
yeah currently we restrict *all* actions to ocfroot, but i only set that up because i was in a hurry

we could probably loosen that up a bit to allow build triggers and maybe some other relatively non-destructive things

parameterized build triggers could still be dangerous i guess, but probably fine

ocflib/account/validators.py
```diff
@@ -5,9 +5,268 @@
 
 import cracklib
 
-import ocflib.constants as constants
 import ocflib.misc.mail
 
+RESERVED_USERNAMES = frozenset((
```
chriskuehl: we should be able to delete this one from ocflib.constants at the same time, right? (given that we already checked and saw that nothing else is using it)

i'm not sure whether this is a great place for this list or not. it might make sense to move elsewhere just because it's so huge. it's probably fine for now.

chriskuehl:
otherwise lgtm, thanks for doing this!

ocflib/account/validators.py
MrMinos: alright. it is done, thank you for the review

tahabi:
should've checked my ocfroot privilege

PR #57, opened 2017-03-11
added time til open

NickImpicciche:


ocflib/lab/hours.py
```diff
@@ -78,6 +78,22 @@ def is_open(self, when=None):
...
+            when = datetime.now()
+
+        if not isinstance(when, datetime):
+            raise ValueError('{} must be a datetime instance'.format(when))
+
+        if self.is_open(when=when):
+            return 0
+        opens = [datetime.combine(self.date, h.open) for h in self.hours]
+        opens = [o for o in opens if o > when]
+        return int((opens[0] - when).total_seconds())
```
chriskuehl: Cool idea. I think as written this will fail if the lab has closed and won't open again today? I guess ideally it'd find the next day we *do* have hours, but that might be difficult over breaks.

I think this should return a `timedelta` instead of seconds, and then you can implement it a bit more simply (instead of `int((opens[0] - when).total_seconds())` just `opens[0] - when`).

ocflib/account/creation.py
```diff
@@ -27,7 +27,7 @@
 from ocflib.printing.quota import SEMESTERLY_QUOTA
 
 
-_KNOWN_UID = 43000
+_KNOWN_UID = 50000
```
jvperrin: Why did you change this to 50000 specifically? The max UID for an account currently is 45530, so this could be bumped to 45500 or something, but we don't have this many accounts yet.

NickImpicciche: this decision was completely arbitrary, wouldn't cutting it that close just mean we have to do this again sooner?

chriskuehl: The "known uid" is the largest known created account. To find the next available UID we have to find the largest already-allocated UID, but LDAP doesn't give us a way to do that efficiently (you have to fetch *all* records and then take the max yourself).

By having a constant of "we know this account exists", we can do an LDAP query like "find all accounts with UID >= knownUID" which lets us only fetch a few hundred or thousand accounts, rather than all 40k+. It's a dumb optimization but we couldn't really come up with a better way.

TL;DR `KNOWN_UID` shouldn't be larger than the current largest allocated UID. @jvperrin's suggestion of 45500 makes sense.

NickImpicciche: okay

matthew-mcallister:
The word is spelled "till"; I just noticed you're missing an "l".

ocflib/lab/hours.py
```diff
@@ -76,8 +77,58 @@ def is_open(self, when=None):
         if not isinstance(when, datetime):
             raise ValueError('{} must be a datetime instance'.format(when))
 
+        if self.date != when.date():
+            raise ValueError('{} is on a different day than {}'.format(when, self))
+
         return any(when in hour for hour in self.hours)
 
+    def time_to_open(self, when=None):
```
NickImpicciche: Is there a reason is_open is done like this? Why not just when=datetime.now()?

chriskuehl: defaults are only evaluated once (when the function is defined), so if you did that, it would never pick up new dates

same reason it's suggested not to use mutable things like dicts or lists as defaults. change them in one invocation of the function, they're changed for all others.

PR #58, opened 2017-03-12
adding semester_dates function

shex1627:
A utility function for the semester job distribution chart. 

1.I thought of making spring_start and fall_start constants but "year" would be a variable that later I have to fix, which I think will make the code harder to understand unless you guys have some better solutions.

### DIFF
--- a/ocflib/lab/stats.py
+++ b/ocflib/lab/stats.py
@@ -1,3 +1,4 @@
+import functools
 from collections import defaultdict
 from collections import namedtuple
 from datetime import date
@@ -72,11 +73,30 @@ def staff_in_lab_count():
 
 
 def current_semester_start():
-    today = date.today()
-    if today.month >= 8:
-        return today.replace(month=8, day=1)
+    return semester_dates()[0]
+
+
+# function partial constants for semesters dates
+_get_spring_start = functools.partial(date, month=1, day=1)
+_get_spring_end = functools.partial(date, month=7, day=31)
+_get_fall_start = functools.partial(date, month=8, day=1)
+_get_fall_end = functools.partial(date, month=12, day=31)
+
+
+def semester_dates(day=None):
+    """Return a tuple (start day, end day) for the current semester.
+
+    Defaults to today if none is provided.
+    """
+    if day is None:
+        day = date.today()
+    fall_start = _get_fall_start(year=day.year)
+    if (day.month, day.day) < (fall_start.month, fall_start.day):
+        return (_get_spring_start(year=day.year),
+                _get_spring_end(year=day.year))
     else:
-        return today.replace(month=1, day=1)
+        return (_get_fall_start(year=day.year),
+                _get_fall_end(year=day.year))
 
 
 def top_staff(start, end=date(3000, 1, 1)):
--- a/tests/lab/stats_test.py
+++ b/tests/lab/stats_test.py
@@ -1,9 +1,11 @@
+from datetime import date
 from datetime import datetime
 from datetime import timedelta
 
 import pytest
 
 from ocflib.lab.stats import list_desktops
+from ocflib.lab.stats import semester_dates
 from ocflib.lab.stats import UtilizationProfile
 
 
@@ -48,3 +50,9 @@ def test_fast_slow_profiles_same():
         assert slow.start == fast.start
         assert slow.end == fast.end
         assert slow.sessions == fast.sessions
+
+
+def test_semester_dates():
+    assert semester_dates() is not None
+    assert semester_dates(date(2016, 1, 1)) == (date(2016, 1, 1), date(2016, 7, 31))
+    assert semester_dates(date(2017, 12, 30)) == (date(2017, 8, 1), date(2017, 12, 31))


shex1627:
So the logic is:
from 01/01 to 07/31 is considered "spring"
and  08/01 to 12/31 is considered "fall"

Basically you have one day, and you expand that day to the semester start and semester end. Subtracting one is to make sure the fall and spring don't overlap.

tahabi:
Oh ok that makes sense

ocflib/lab/stats.py
```diff
@@ -79,6 +79,22 @@ def current_semester_start():
...
+    one_day = timedelta(days=1)
+    year = datetime.today().year
+    if day is None:
+        day = date.today()
+    spring_start = date(year, 1, 1)
+    fall_start = date(year, 8, 1)
+    if day.month < 8:
+        return (spring_start, fall_start - one_day)
+    else:
+        return (fall_start, spring_start - one_day)
```
kpengboy: Shouldn't this be the `spring_start` of the next year

ocflib/lab/stats.py
```diff
@@ -79,6 +79,22 @@ def current_semester_start():
...
+    """Return a tuple (start day, end day) for the current semester.
+    Defaults to today if none is provided.
+    """
+    one_day = timedelta(days=1)
+    year = datetime.today().year
+    if day is None:
+        day = date.today()
+    spring_start = date(year, 1, 1)
+    fall_start = date(year, 8, 1)
+    if day.month < 8:
```
kpengboy: I would've written it `if day < fall_start`, I think that's more intuitive...

ocflib/lab/stats.py
shex1627: this is part of the SQLquery for the jobs, so not exactly the academic year.

ocflib/lab/stats.py
shex1627: tested, and it works. Originally I was concerned that we can't compare day with fall_start and I was lazy to figure that out.

ocflib/lab/stats.py
shex1627: sorry Kevin I was wrong. Do you have any better way of fixing this? 
My current proposal is
```python
if day < fall_start:                                           
    return (spring_start, fall_start - ONE_DAY)                  
else:                                                              
    return (fall_start, spring_start.replace(year=day.year+1) - ONE_DAY)
```

ocflib/lab/stats.py
```diff
@@ -79,6 +79,22 @@ def current_semester_start():
         return today.replace(month=1, day=1)
 
 
+def semester_dates(day=None):
+    """Return a tuple (start day, end day) for the current semester.
+    Defaults to today if none is provided.
```
matthew-mcallister: Put a newline before this line.

ocflib/lab/stats.py
```diff
@@ -79,6 +79,22 @@ def current_semester_start():
...
+    """
+    if day is None:
+        day = date.today()
+    year = day.year
+    spring_start = date(year, 1, 1)
+    fall_start = date(year, 8, 1)
+    fall_end = date(year, 12, 31)
+    if day < fall_start:
```
matthew-mcallister: This will error if you give it a `datetime` object. Can you replace this with something like
```
if date(day.year, day.month, day.day) < fall_start:
```
to implicitly convert `datetime` to `date`?

shex1627:
```python
if day.month < fall_start.month:
```
?

ocflib/lab/stats.py
shex1627: which line(number)?

ocflib/lab/stats.py
chriskuehl: I'd rather not assume it starts/ends on a month if we can avoid it, though that is currently the case.

ocflib/lab/stats.py
matthew-mcallister: Should be
```
"""Return a tuple (start day, end day) for the current semester.

Defaults to today if none is provided.
"""
```

ocflib/lab/stats.py
matthew-mcallister: Good try, but I think @chriskuehl is right.

matthew-mcallister:
Also, can you add a simple test for this? Not really sure what would be good. Maybe just pick a couple dates that you think should be in the same semester and test that they return the same tuple?

shex1627:
commented so that you guys can see the new commit.

tests/lab/stats_test.py
```diff
@@ -48,3 +50,9 @@ def test_fast_slow_profiles_same():
         assert slow.start == fast.start
         assert slow.end == fast.end
         assert slow.sessions == fast.sessions
+
+
+def test_semester_dates():
+    assert semester_dates() is not None
+    assert semester_dates(date(2016, 1, 1)) == (date(2016, 1, 1), date(2016, 7, 31))
+    assert semester_dates(date(2017, 12, 30)) == (date(2017, 8, 1), date(2017, 12, 31))
```
matthew-mcallister: Thanks for the tests.

ocflib/lab/stats.py
```diff
@@ -79,6 +83,22 @@ def current_semester_start():
...
+    """Return a tuple (start day, end day) for the current semester.
+
+    Defaults to today if none is provided.
+    """
+    if day is None:
+        day = date.today()
+    year = [day.year]
+    if [day.month, day.year] < FALL_START:
+        return (date(*(year + SPRING_START)),
+                date(*(year + SPRING_END)))
```
matthew-mcallister: Why all this list arithmetic? Just make `SPRING_START` etc. namedtuples or dates and do
```
return (date(year, SPRING_START.month, SPRING_START.day), date(year, SPRING_END.month, SPRING_END.day))
```

shex1627: If I use python Date object then I have to use a dummy attribute 'year' since teach year is difference.
```python
#maybe we will use that in the future
DUMMY_YEAR = 1000
SPRING_START = date(DUMMY_YEAR, 1, 1)
# ...
return (date(year, SPRING_START.month, SPRING_START.day), 
date(year, SPRING_END.month, SPRING_END.day))
```

if I use namedtuples then any other files that use those constants may have to import the 'collection' library.
```python
from collections import namedtuple
Date_No_Year <- namedtuple('month', 'day')
SPRING_START = Date_No_Year(1, 1)
```
Do you like the dummy_year implementation better or creating a new structure using namedtuple?

kpengboy: I'd say pick one of them and go with it. Both approaches have their own drawbacks.

If you go with the dummy attribute route, note that you can use the `replace` method on a `date` object to simplify your code.

If you create a new namedtuple, note that you can do something like this:

    return (date(year, *SPRING_START), date(year, *SPRING_END))

ocflib/lab/stats.py
```diff
@@ -79,6 +83,22 @@ def current_semester_start():
         return today.replace(month=1, day=1)
 
 
+def semester_dates(day=None):
+    """Return a tuple (start day, end day) for the current semester.
```
kpengboy: You probably meant "for the semester containing the given day"?

ocflib/lab/stats.py
```diff
@@ -79,6 +83,22 @@ def current_semester_start():
...
+
+    Defaults to today if none is provided.
+    """
+    if day is None:
+        day = date.today()
+    year = [day.year]
+    if [day.month, day.year] < FALL_START:
```
kpengboy: I think you mean `day.day` instead of `day.year`?

kpengboy:
While you're at it, you can rewrite the `current_semester_start()` function to call `semester_dates` instead.

shex1627:
New revision updated. Why is Kevin so good at detecting logical bugs?

shex1627:
could anyone have a look at this?

ocflib/lab/stats.py
```diff
@@ -13,6 +13,14 @@
 # when we started keeping stats
 STATS_EPOCH = date(2014, 2, 15)
 
+# constants for semesters dates
+# dummy year variable
+DUMMY_YEAR = 1000
```
matthew-mcallister: I don't think there's any reason to name this constant (and superfluous constants pollute the namespace). How about just leaving a comment? For example,
```
# The year is a placeholder; we only care about month and day.
SPRING_START = date(1, 1, 1)
```

ocflib/lab/stats.py
```diff
@@ -13,6 +13,14 @@
...
+SPRING_START = date(DUMMY_YEAR, 1, 1)
+SPRING_END = date(DUMMY_YEAR, 7, 31)
+FALL_START = date(DUMMY_YEAR, 8, 1)
+FALL_END = date(DUMMY_YEAR, 12, 31)
```
chriskuehl: this feels really weird, what about just doing

```python
import functools
from datetime import date

spring_start = functools.partial(date, month=1, day=1)
spring_end = functools.partial(date, month=7, day=31)
fall_start = functools.partial(date, month=8, day=1)
fall_end = functools.partial(date,month=12, day=31)
```

then later instead of doing `SPRING_START.replace(year=2017)` just do `spring_start(year=2017)`

shex1627:
@chriskuehl I use your functool solution, but also rename constants to be like FALL_START_FUN so that others know they are not date objects when they want to use them.

ocflib/lab/stats.py
```diff
@@ -72,11 +79,23 @@ def staff_in_lab_count():
...
+    """
+    if day is None:
+        day = date.today()
+    fall_start = FALL_START_FUNP(year=day.year)
+    if (day.month, day.day) < (fall_start.month, fall_start.day):
+        return (SPRING_START_FUNP(year=day.year),
+                SPRING_END_FUNP(year=day.year))
     else:
-        return today.replace(month=1, day=1)
+        return (FALL_START_FUNP(year=day.year),
```
matthew-mcallister: This could be `fall_start` but it doesn't really matter.

ocflib/lab/stats.py
```diff
@@ -13,6 +14,12 @@
 # when we started keeping stats
 STATS_EPOCH = date(2014, 2, 15)
 
+# function partial constants for semesters dates
+SPRING_START_FUNP = functools.partial(date, month=1, day=1)
+SPRING_END_FUNP = functools.partial(date, month=7, day=31)
+FALL_START_FUNP = functools.partial(date, month=8, day=1)
+FALL_END_FUNP = functools.partial(date, month=12, day=31)
```
chriskuehl: these are functions, why are they uppercase? they're not constants.

you don't need the `FUNP` suffix to mark them as functions if you just name them like functions :)

shex1627: although those are functions but I think the names could be misleading for people to think they are date objects. Do we need to do anything about that? Or You don't think it is a problem at all.

tahabi:
rename them entirely to something more indicative of being functions like
`get_fall_start` or something like that?

On Apr 30, 2017 3:23 PM, "shicheng" <notifications@github.com> wrote:

*@shex1627* commented on this pull request.
------------------------------

In ocflib/lab/stats.py
<https://github.com/ocf/ocflib/pull/58#discussion_r114085819>:

> @@ -13,6 +14,12 @@
 # when we started keeping stats
 STATS_EPOCH = date(2014, 2, 15)

+# function partial constants for semesters dates
+SPRING_START_FUNP = functools.partial(date, month=1, day=1)
+SPRING_END_FUNP = functools.partial(date, month=7, day=31)
+FALL_START_FUNP = functools.partial(date, month=8, day=1)
+FALL_END_FUNP = functools.partial(date, month=12, day=31)

although those are functions but I think the names could be misleading for
people to think they are date objects. Do we need to do anything about
that? Or You don't think it is a problem at all.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/ocf/ocflib/pull/58#discussion_r114085819>, or mute the
thread

ocflib/lab/stats.py
matthew-mcallister: I agree the name should indicate that these objects are callable. How about using a verb in the name? E.g. `get_fall_start` or even `fall_start_for_year`.

shex1627: those functions are not callable but simply a function that takes the year as argument to be a date object. I think `fall_start_funp` is a decent option, at least people would get confused and lookup where the variables are defined. 

By that standard, the dummy-year-date-constant implementation is not a bad alternative.

shex1627:
renaming all those function partials to get_xxx_xxx.

ocflib/lab/stats.py
```diff
@@ -13,6 +14,12 @@
 # when we started keeping stats
 STATS_EPOCH = date(2014, 2, 15)
 
+# function partial constants for semesters dates
+get_spring_start = functools.partial(date, month=1, day=1)
+get_spring_end = functools.partial(date, month=7, day=31)
+get_fall_start = functools.partial(date, month=8, day=1)
+get_fall_end = functools.partial(date, month=12, day=31)
```
matthew-mcallister: Now that I think about it, these should probably be prefixed with `_`.

shex1627:
adding prefix "_" to the functions

shex1627:
do you guys want me to squash the commits as well

matthew-mcallister:
Oh, it looks like the squash and merge button isn't available. You can if you want, otherwise I will.

shex1627:
I will mess with my local repo see if I can do it myself first. Can't rely on others every time

shex1627:
Seems like I did it(squash the commits) guys
Not sure if the standard way
first 
git reset --soft HEAD~9 && git commit
git push --force origin semesterJobUtil

matthew-mcallister:
shipit

PR #61, opened 2017-03-21
Add app hosting module

matthew-mcallister:
It's pretty much a clone of vhost.web, admittedly. I thought about sharing code between modules, but it seemed more complicated than it was worth for such a small amount of code.

### DIFF
--- a/ocflib/vhost/application.py
+++ b/ocflib/vhost/application.py
@@ -1,0 +1,51 @@
+import requests
+
+VHOST_DB_PATH = '/home/s/st/staff/vhost/vhost-app.conf'
+VHOST_DB_URL = 'https://www.ocf.berkeley.edu/~staff/vhost-app.conf'
+
+
+def get_app_vhost_db():
+    """Returns lines from the application vhost database. Loaded from the
+    filesystem (if available), or from the web if not."""
+    try:
+        with open(VHOST_DB_PATH) as f:
+            return list(map(str.strip, f))
+    except IOError:
+        # fallback to database loaded from web
+        return requests.get(VHOST_DB_URL).text.split('\n')
+
+
+def get_app_vhosts():
+    """Returns a list of application virtual hosts in convenient format.
+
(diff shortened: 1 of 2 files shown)


ocflib/vhost/application.py
```diff
@@ -0,0 +1,51 @@
...
+VHOST_DB_PATH = '/home/s/st/staff/vhost/vhost-app.conf'
+VHOST_DB_URL = 'https://www.ocf.berkeley.edu/~staff/vhost-app.conf'
+
+
+def get_app_vhost_db():
+    """Returns lines from the application vhost database. Loaded from the
+    filesystem (if available), or from the web if not."""
+    try:
+        with open(VHOST_DB_PATH) as f:
+            return list(map(str.strip, f))
```
chriskuehl: lol i know i wrote this but to be more accurate to the thing below, it should probably be `return f.read().splitlines()`

ocflib/vhost/application.py
```diff
@@ -0,0 +1,51 @@
...
+            continue
+
+        fields = line.split(' ')
+
+        username, host, socket, ssl_cert = fields
+
+        vhosts[fully_qualify(username if host == '-' else host)] = {
+            'username': username,
+            'socket': socket if socket != '-' else username,
+            'ssl_cert': ssl_cert if ssl_cert != '-' else None,
```
chriskuehl: eventually we'll probably want to remove this once we have let's encrypt (since everything will have an ssl cert, there will be no need to specify them in the config file)

matthew-mcallister: Yeah, I'm taking it one step at a time for now since I want to have this available in ocflib to start updating the apphost server.

ocflib/vhost/application.py
matthew-mcallister: Sure, it's not like the whitespace is hurting anything. Will get that in vhost.web too.

PR #63, opened 2017-04-07
move create pubkey to __init__ from constants

tahabi:
this is used in ocfweb/account/register.py and approve, pushing those PRs as well


matthew-mcallister:
High quality commit message.

tahabi:
lol

better?

matthew-mcallister:
Meh, marginally.

tahabi:
care to suggest a better one?

jvperrin:
I take that back on the explicit import part, since it still is, but I do think it would be clearer if it wasn't in `__init__` since we don't tend to put constants there and the public key is used for account creation anyway.

tahabi:
It seemed cleaner to put it in __init__ but I don't have any particularly strong opinions on the topic

chriskuehl:
i agree, imo the only thing that should ever go in `__init__.py` is imports (and then only when you're trying to create an explicit public API, the way `ldap3` or `requests` does it)

people don't expect to find code in `__init__.py` which is a good reason not to put it there

PR #64, opened 2017-04-08
Use correct timezone

kpengboy:
The Z stands for UTC time, which is misleading because our times are not
actually in UTC time. This fixes it to show the correct timezone.


matthew-mcallister:
Wow, I didn't even realize. If it works, ship it.

kpengboy:
Looks like I've caught all the places which need to be updated to produce an aware datetime object. If I've missed any, we'll get a `ValueError` soon and fix that...

PR #66, opened 2017-04-28
changing hours yolo

douglaslwong:


ocflib/lab/hours.py
```diff
@@ -149,6 +149,7 @@ def closed_all_day(self):
     (date(2017, 3, 24), date(2017, 3, 24), 'Friday before Spring Break', [Hour(time(9), time(12))]),
     (date(2017, 3, 25), date(2017, 4, 2), 'Spring Break', []),
     (date(2017, 4, 27), date(2017, 4, 27), 'MLK is closed', []),
+    (date(2017, 4, 28), date(2017, 4, 28), 'Yash Day', [Hour(time(14), time(16))]),
```
chriskuehl: ideally this would say "early close" or something since this is visible to users

chriskuehl: to be actually correct the hours should probably be `[Hour(time(9, 10), time(11)), Hour(time(14, 10), time(16))]` but i guess it's not the worst given that we're already past 2pm

douglaslwong:
thumbs up to myself, merging

ocflib/lab/hours.py
douglaslwong: jk reverting this

jvperrin:
Why not just fix it then merge instead of having to revert again??

wat

douglaslwong:
nickimp is working on it already, sorry about that

jvperrin:
Haha, it's fine, I'm just very confused

douglaslwong:
lol ended up being andrew, but yeah doing this before going into a project meeting was a bad idea

PR #69, opened 2017-05-01
update hours for dead week

tahabi:
not fun trying to do this on the phone so let me know if I've made any mistakes


asaiacai:
From what I saw, it all checks out!

PR #70, opened 2017-05-01
pre-emptive finals week hours

tahabi:


chriskuehl:
I think a better way to do this would have been to add these days as holidays. This way, the "next week's hours" would work correctly.

NickImpicciche:
This actually won't even pass test because you can only enter in empty days through holidays.

Definitely change this to holidays.

PR #71, opened 2017-05-20
add mirrors SQL schema

tahabi:
not sure where to put this since the rest of ocfstats is documented in labstats but all the functions using this are in ocfweb

### DIFF
--- a/ocflib/lab/mirrors.sql
+++ b/ocflib/lab/mirrors.sql
@@ -1,0 +1,16 @@
+---
+--- documentation of `mirrors` table in ocfstats
+--- other tables are defined in labstats/labstats/bin/init.py
+---
+
+CREATE TABLE `mirrors` (
+  `date` date NOT NULL,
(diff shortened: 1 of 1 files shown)


ocflib/lab/mirrors.sql
```diff
@@ -0,0 +1,16 @@
...
+---
+
+CREATE TABLE `mirrors` (
+  `date` date NOT NULL,
+  `dist` varchar(30) NOT NULL,
+  `up` bigint(20) NOT NULL,
+  `down` bigint(20) NOT NULL
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+CREATE VIEW `mirrors_public` AS
```
tahabi: superfluous but done to maintain consistency with granting select to anonymous on public views only

chriskuehl: seems reasonable to me, explicit is better than implicit and all that other zen

PR #72, opened 2017-05-20
Add vhost.config for generating server configs

matthew-mcallister:
As part of this addition, the app vhost config file now uses the same format as the regular vhost config file.

The corresponding changes to puppet will have to wait until revisions here are done.

### DIFF
--- a/ocflib/vhost/config.py
+++ b/ocflib/vhost/config.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+import os
+from collections import namedtuple
+from pathlib import Path
+
+import jinja2
+
+from ocflib.account.search import user_is_sorried
+from ocflib.misc.mail import email_for_user
+from ocflib.vhost.application import get_app_vhosts
+from ocflib.vhost.web import get_vhosts
+
+
+SITE_CFG = '/etc/nginx/sites-enabled/virtual'
+
+LETS_ENCRYPT_SSL = Path('/services/http/ssl')
+SSL_BUNDLE = Path('/etc/ssl/apphost')
+
+APP_DIR = Path('/srv/apps')
+WEB_DIR = Path('/services/http/users')
+
+jinja_env = jinja2.Environment(
+    loader=jinja2.FileSystemLoader((
+        os.path.abspath(os.path.dirname(__file__)),
+    )),
+)
+
+
+class VirtualHost(namedtuple('VirtualHost', (
+    'fqdn',
+    'user',
+    'socket_or_docroot',
+    'aliases',
+    'flags',
+))):
+    """A single virtual host as defined by the vhost config files.
+
+    This class defines lots of helper functions which are used in
+    generating vhost configuration on the web and app hosts.
+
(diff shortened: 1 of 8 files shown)


ocflib/vhost/config.py
```diff
@@ -0,0 +1,136 @@
...
+
+import jinja2
+
+from ocflib.account.search import user_is_sorried
+from ocflib.misc.mail import email_for_user
+from ocflib.vhost.application import get_app_vhosts
+from ocflib.vhost.web import get_vhosts
+
+
+SITE_CFG = '/etc/nginx/sites-enabled/virtual'
```
chriskuehl: i don't think ocflib show know anything about nginx... that seems like strictly application territory, not library

ocflib/vhost/config.py
```diff
@@ -0,0 +1,136 @@
...
+    def ssl_chain(self):
+        return '/etc/ssl/certs/lets-encrypt.crt'
+
+    @property
+    def ssl_bundle(self):
+        return str(SSL_BUNDLE / (self.fqdn + '.crt'))
+
+
+def build_web_config():
+    """Generates an apache config file for web vhosts from the config
```
chriskuehl: i really feel this stuff doesn't belong in ocflib...

ocflib/vhost/vhost-app.jinja
```diff
@@ -0,0 +1,51 @@
...
+        {%- for alias in vhost.aliases %} "{{alias}}"{% endfor %};
+
+    location /.well-known/ {
+        proxy_pass https://www.ocf.berkeley.edu/.well-known/;
+    }
+
+    location / {
+        {% if vhost.disabled %}
+            # Proxy to the "unavailable" site, just as regular vhosts do.
+            proxy_pass https://unavailable.ocf.berkeley.edu/;
```
chriskuehl: does https actually work here? the cert is invalid...

ocflib/vhost/vhost-app.jinja
```diff
@@ -0,0 +1,51 @@
...
+
+    {% if vhost.use_ssl %}
+        ssl on;
+        ssl_certificate {{vhost.ssl_bundle}};
+        ssl_certificate_key {{vhost.ssl_key}};
+        add_header Strict-Transport-Security "max-age=31536000";
+    {% endif %}
+}
+{% if vhost.use_ssl %}
+server {
```
chriskuehl: indent?

ocflib/vhost/config.py
matthew-mcallister: This particular line was actually included by mistake.

matthew-mcallister:
I don't remember all the details of why I originally had the idea to put all of this code in ocflib (something about deploying death on mesos), but that doesn't matter because ultimately @chriskuehl is right that most of it doesn't belong here.

For what it's worth, I don't think it would be too much of a stretch to have ocflib generate generic config files. Stuff like fastcgi and php in the apache config template, though, are way too specific to the host config to be here.

PR #73, opened 2017-05-27
Converge vhost config formats

matthew-mcallister:
Since web applications should be getting SSL certs from Let's Encrypt,
this changes the last field in the app vhost config to contain a list of
alias domains where it previously had the name of the website's SSL
cert. This also removes the now-unused "!" syntax from the vhost file.

### DIFF
--- a/ocflib/vhost/application.py
+++ b/ocflib/vhost/application.py
@@ -1,3 +1,5 @@
+import re
+
 import requests
 
 VHOST_DB_PATH = '/home/s/st/staff/vhost/vhost-app.conf'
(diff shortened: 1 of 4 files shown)


ocflib/vhost/application.py
```diff
@@ -42,12 +45,18 @@ def fully_qualify(host):
 
         fields = line.split(' ')
 
-        username, host, socket, ssl_cert = fields
+        if len(fields) < 5:
+            flags = []
+        else:
+            flags = re.search('^\[(.*)\]$', fields[4]).group(1).split(',')
```
chriskuehl: might as well use `re.match` if you're going to anchor to the start anyway

matthew-mcallister: I just copied from `vhost.web`; I can change it in both places if it seems appropriate.

jvperrin:
:clap: :clap: :clap: :clap: :clap:

PR #74, opened 2017-07-11
Big Brother

asaiacai:
Migrating OCF hours to YAML file so that we don't have to rebuild to change hours.
Example YAML file: https://gist.github.com/asaiacai/ddeda8c5091739fbd5ca8b47048fe915
Tested on previous hours we had: https://docs.google.com/document/d/1LatfQZjGXPzWYcyMrE3SwuFPLbrcuNue5cGEkdbnRw8/edit
Example output:
https://gyazo.com/502d1e674b7be660819403090ac66bb6

### DIFF
--- a/ocflib/lab/hours.py
+++ b/ocflib/lab/hours.py
@@ -20,10 +20,18 @@ from datetime import datetime
 from datetime import time
 from datetime import timedelta
 
+import requests
+
+SHIFT_LENGTH = timedelta(hours=1)
+SHIFTS_URL = 'https://docs.google.com/spreadsheet/ccc?key=1WgczUrxqey63fmPRmdkCDMCbsfaTyCJrixiCeJj35UI&output=csv'
+DAY_OFFSET = 1
+
 
 class Hour(namedtuple('Hours', ['open', 'close'])):
 
     def __contains__(self, when):
+        if isinstance(when, time):
+            return self.open <= when < self.close
         return self.open <= when.time() < self.close
 
 
@@ -132,6 +140,88 @@ class Day(namedtuple('Day', ['date', 'weekday', 'holiday', 'hours'])):
     @property
     def closed_all_day(self):
         return not self.hours
+
+
+def _shift_from_start(start):
+    """Returns an Hour object from the start time"""
+    end = (datetime.combine(date.today(), start) + SHIFT_LENGTH).time()
+    return Hour(start, end)
+
+
+def _string_to_time(when):
+    return datetime.strptime(when, '%I:%M %p').time()
+
+
+def _shift_matrix():
+    """Returns the OCF staff shifts from google spreadsheet
+    as a 2-D matrix"""
+    response = requests.get(SHIFTS_URL)
+    content = response.content.decode('utf-8').splitlines()
+    return [row.split(',') for row in content]
+
+
+def get_shifts(when=None):
+    """Fetches lab shifts for given datetime instance or weekday int."""
+    if when is None:
+        when = datetime.now()
+
+    shifts = _shift_matrix()[1:]
+    try:
+        return [_shift_from_start(_string_to_time(hour[0]))
+                for hour in shifts if hour[when.weekday() + DAY_OFFSET]]
+    except AttributeError:
+        return [_shift_from_start(_string_to_time(hour[0]))
+                for hour in shifts if hour[when + DAY_OFFSET]]
+
+
+def union(shift1, shift2):
+    """
+    Returns the combined Hour if there is an overlap,
+    otherwise, return None
+    """
+    if shift1.close in shift2 or shift2.close in shift1:
+        return Hour(min(shift1.open, shift1.open),
+                    max(shift1.close, shift2.close))
+    else:
+        return None
+
+
+def get_hours(when=None):
+    """Combines lab shifts as much as possible and returns the final list of Hours for
+    'day'.  'day' is an int [0,6] or datetime object"""
+    if when is None:
+        when = datetime.now()
+
+    shifts = get_shifts(when)
+    i = 0
+    while i < len(shifts):
+        j = i + 1
+        while j < len(shifts):
+            hour = union(shifts[i], shifts[j])
+            if hour is not None:
+                shifts.pop(j)
+                shifts[i] = hour
+            else:
+                j += 1
+        i += 1
+    return shifts
+
+
+def staff_on_shift(when=None):
+    """Finds the staffer on shift during 'when' according to
+    google spreadsheet.  Returns None if no staffer found.
+    'when' is a datetime object.
+
+    Does not support holiday hours/shifts.
+    """
+    if when is None:
+        when = datetime.now()
+
+    shifts = _shift_matrix()[1:]
+    for hour in shifts:
+        if when in _shift_from_start(
+                _string_to_time(hour[0])):
+            return hour[when.weekday() + DAY_OFFSET]
 
 
 REGULAR_HOURS = defaultdict(lambda: [Hour(time(9), time(19))], {


ocflib/lab/hours.py
```diff
@@ -20,11 +20,18 @@
 from datetime import time
 from datetime import timedelta
 
+from yaml import load
```
chriskuehl: you always want `safe_load`, `yaml.load` is extremely dangerous:

```python
>>> import yaml
>>> yaml.load("!!python/object/apply:os.system\nargs: ['echo pwned']")
pwned
0
```

just a personal style thing, but i'd also import this as `import yaml` rather than `from yaml import load`. it's a little clearer later in the code if you do `yaml.load(...)` rather than `load(...)` as to what kind of loading you're actually doing

ocflib/lab/stats.py
```diff
@@ -65,6 +65,17 @@ def staff_in_lab():
         return [Session.from_row(r) for r in c]
 
 
+def frontdesk_logged():
+    """Returns whether the front desktop is logged in"""
+    staff_in_lab = None
+
+    staff_in_lab = staff_in_lab()
+    for staff in staff_in_lab:
+        if staff.host == 'frontdesk.ocf.berkeley.edu':
```
chriskuehl: are we actually naming the host frontdesk? seems like that should be a CNAME for a disaster-named host instead.

ocflib/lab/hours.py
```diff
@@ -134,14 +141,91 @@ def closed_all_day(self):
...
-    Day.SATURDAY: [Hour(time(11, 10), time(15))],
-    Day.SUNDAY: [Hour(time(11, 10), time(15))],
+    Day.MONDAY: get_hours(Day.MONDAY),
+    Day.TUESDAY: get_hours(Day.TUESDAY),
+    Day.WEDNESDAY: get_hours(Day.WEDNESDAY),
+    Day.THURSDAY: get_hours(Day.THURSDAY),
+    Day.FRIDAY: get_hours(Day.FRIDAY),
+    Day.SATURDAY: get_hours(Day.SATURDAY),
+    Day.SUNDAY: get_hours(Day.SUNDAY),
 })
```
chriskuehl: i wonder if this actually helps all that much. as written, you're still going to need to re-deploy ocfweb if you want the new hours to take effect, so why store them in a yaml file anyway?

ocflib/lab/hours.py
```diff
@@ -134,14 +141,91 @@ def closed_all_day(self):
...
+    return datetime.strptime(when, '%I:%M%p').time()
+
+
+def get_shifts(when=None):
+    """Fetches lab shifts for given datetime instance or weekday int."""
+    if not when:
+        when = datetime.now()
+
+    hours = []
+    stream = open(FRONTDESK_HOURS_FILE)
```
chriskuehl: use a context manager so the file gets closed explicitly

```python
with open(FRONTDESK_HOURS_FILE) as f:
    data = yaml.safe_load(f)
```

ocflib/lab/hours.py
```diff
@@ -134,14 +141,91 @@ def closed_all_day(self):
...
+    data = load(stream)
```
chriskuehl: you're going to need some kind of fallback, NFS isn't mounted on most hosts (including ocfweb)

in other places, we fall back by reading the file via http from death, maybe that's an option?

ocflib/lab/hours.py
```diff
@@ -134,14 +141,91 @@ def closed_all_day(self):
         return not self.hours
 
 
+def string_to_time(when=None):
```
chriskuehl: prefix with an underscore since we don't want consumers using this method

ocflib/lab/hours.py
```diff
@@ -134,14 +141,91 @@ def closed_all_day(self):
...
+    regular hours in hours.yaml.  Returns None if no staffer found.
+    'when' is a datetime object.
+
+    Does not support holiday hours/shifts.
+    """
+
+    if not when:
+        when = datetime.now()
+
+    data = load(open(FRONTDESK_HOURS_FILE))
```
chriskuehl: same comments here

tahabi:
I think a better design for the YAML file would be day : hour : person on shift, rather than person on shift : day : hour. Currently, days are implicit and there are empty blocks on days when someone isn't scheduled, plus I think the alternative design may make your logic simpler, for example, when trying to find who is on shift for a particular hour.

ocflib/lab/stats.py
jvperrin: I'm not actually sure what the best way to find which desktop is at the front desk, maybe pulling from a hiera setting?

ocflib/lab/hours.py
jvperrin: Ah, that's a good point, I didn't catch that when I read through this before, but you're right that the calls to `get_hours` would only be run once on startup and then a new deploy would need to be made, so changing this to be more dynamic would be good.

ocflib/lab/hours.py
jvperrin: For example, [`_load_staff_hours`](https://github.com/ocf/ocflib/blob/17bfaa055867c288920ad8316be512144f132065/ocflib/lab/staff_hours.py#L32-L39)

ocflib/lab/stats.py
```diff
@@ -65,6 +65,17 @@ def staff_in_lab():
         return [Session.from_row(r) for r in c]
 
 
+def frontdesk_logged():
+    """Returns whether the front desktop is logged in"""
+    staff_in_lab = None
```
asaiacai: I think that line didn't pass precommit when I was just doing the function call. Yeah i thought that was weird too

asaiacai:
@jvperrin I agree that a spreadsheet is probably easier to proofread and more accessible.  Maybe having it reference a Google sheet in the OCF drive will be easier to edit and less error prone for other staff than having to SSH and edit a YAML file.  We can have the columns be days and rows be time slots.

chriskuehl:
lol yeah, we definitely can't have non-technical people editing a YAML file, especially in a terminal text editor

our own staff can barely edit a YAML file correctly :P

jvperrin:
Haha yeah, it's very easy to mess up with YAML given the indentation requirements and all that, and I think having a spreadsheet will definitely be more familiar anyway. I've created an `ocfbot@ocf.b.e` account to be used with APIs and stuff, so I'll try and look into Google Sheets APIs and see what is possible there.

ocflib/lab/stats.py
matthew-mcallister: Can't you check for the frontdesk CNAME in the host's LDAP entry?

ocflib/lab/hours.py
```diff
@@ -134,14 +141,91 @@ def closed_all_day(self):
...
+        i += 1
+    return shifts
+
+
+def staff_on_shift(when=None):
+    """Finds the staffer on shift during 'when' according to
+    regular hours in hours.yaml.  Returns None if no staffer found.
+    'when' is a datetime object.
+
+    Does not support holiday hours/shifts.
```
matthew-mcallister: How do you actually plan to support holiday shifts? Just add an extra lookup table for each holiday into the YAML file?

matthew-mcallister:
If the Google API is insufficient, we can just ask the OM to export it to CSV and make some kind of uploader frontend on the website or as a script.

ocflib/lab/stats.py
jvperrin: Yes, good point, that's much better

chriskuehl: yeah, either searching for dnsCname in ldap or doing an actual dns lookup sound good

ocflib/lab/hours.py
tahabi: we could implement this the same way we implement staff hours right? which doesn't need ocfweb to be redeployed in order to update, iirc?

asaiacai:
Ok so i updated the program to use a google spreadsheet (linked below)

https://docs.google.com/a/berkeley.edu/spreadsheets/d/1WgczUrxqey63fmPRmdkCDMCbsfaTyCJrixiCeJj35UI/edit?usp=sharing

Sample outputs:
https://gyazo.com/93535c825e4eb5598f4498d703952b46
https://gyazo.com/93535c825e4eb5598f4498d703952b46

For the dynamic hours, I was under the impression that we don't have to redeploy in the case of staff hours.  Could we do something similar in this case as @tahabi mentioned?

ocflib/lab/gsheet.py
```diff
@@ -0,0 +1,75 @@
+from datetime import datetime
```
tahabi: I'd move this entire file to like `ocflib.misc.gsheet`

setup.py
```diff
@@ -24,8 +24,10 @@
         'cached_property',
         'cracklib',
         'dnspython',
+        'gspread',
         'jinja2',
         'ldap3',
+        'oauth2client',
```
chriskuehl: I'm kinda sad to be adding these as dependencies to ocflib.

I think it'd be better for this code to live in an application (e.g. ocfweb) and provide an API which others can call, especially since it requires special permissions and configuration.

ocflib has too many dependencies as it is. Additionally, gspread doesn't appear to be Debian-packaged, so we can't add it as a dependency unless we want to package it ourselves for both jessie and stretch.

ocflib/lab/gsheet.py
```diff
@@ -0,0 +1,75 @@
...
+def shifts_sheet():
+    """Returns the worksheet containing the OCF
+    frontdesk shifts"""
+    return get_sheet(SHEET_NAME, SHIFTS_JSON_PATH)
+
+
+def _day_to_col(when):
+    try:
+        return when + DAY_COLUMN_OFFSET
+    except:
```
chriskuehl: don't use blank except, at least it should be `except Exception` so that it doesn't catch stuff like KeyboardInterrupt or SystemExit (but really it should catch whatever you're expecting this to raise)

ocflib/lab/gsheet.py
```diff
@@ -0,0 +1,75 @@
...
+
+    hours = _row_to_start()
+    wk = shifts_sheet()
+    todays_hours = []
+    for hour in hours:
+        if wk.cell(hours[hour], _day_to_col(when)).value:
+            todays_hours += [hour]
+    return todays_hours
+
+    return todays_hours
```
chriskuehl: duplicate return?

ocflib/lab/gsheet.py
```diff
@@ -0,0 +1,75 @@
...
+
+def _string_to_time(when):
+    return datetime.strptime(when, '%I:%M %p').time()
+
+
+def _row_to_start():
+    """Returns a dictionary mapping start times
+    to row numbers"""
+    i = HOURS_START_OFFSET
+    d = {}
```
chriskuehl: what is d?

ocflib/lab/gsheet.py
```diff
@@ -0,0 +1,75 @@
...
+    wk = shifts_sheet()
+    shift_times = wk.col_values(1)
+    while shift_times[i]:
+        d[_string_to_time(shift_times[i])] = i + 1  # +1 compensate for 0 index
+        i += 1
```
chriskuehl: i'd use enumerate, this doesn't feel very pythonic

```python
for i, shift_time in enumerate(shift_times):
    d[_string_to_time(shift_time)] = i + 1
```

tahabi:
Instead of integrating the spreadsheet-reading functionality into ocflib, I think it would make more sense to write another script that reads the sheet, converts it to YAML, and places that in ~staff. That way, we can run that script as a cronjob every 5 minutes or something on supernova and have ocfweb refresh itself without having to redeploy like staff hours are done already. This way, any staff/opstaff can edit the sheet, technical acumen notwithstanding, and have it be live basically 5 minutes later. Other benefits include not having to add more dependencies to ocflib itself, and perhaps even not needing to share the spreadsheet URL publicly.

chriskuehl:
ocfweb already has the capability of doing cronjob-like things (periodic functions) and is easier to test than a cronjob that dumps stuff onto a special location on NFS.

ocflib/lab/gsheet.py
matthew-mcallister: Or use a comprehension:
```python
return {
    _string_to_time(shift_time): i + 1 + HOURS_OFFSET
    for i, shift_time in enumerate(shift_times)
}
```

ocflib/lab/gsheet.py
```diff
@@ -0,0 +1,75 @@
...
+    or int on [0,6]"""
+    if when is None:
+        when = datetime.now()
+
+    hours = _row_to_start()
+    wk = shifts_sheet()
+    todays_hours = []
+    for hour in hours:
+        if wk.cell(hours[hour], _day_to_col(when)).value:
+            todays_hours += [hour]
```
matthew-mcallister: You could use a comprehension here too if you wanted.

ocflib/lab/hours.py
```diff
@@ -134,14 +142,83 @@ def closed_all_day(self):
...
+    while i < len(shifts):
+        j = i + 1
+        while j < len(shifts):
+            hour = union(shifts[i], shifts[j])
+            if hour is not None:
+                shifts.pop(j)
+                shifts[i] = hour
+            else:
+                j += 1
+        i += 1
```
matthew-mcallister: It might be too much, but you could use a generator instead:
```python
acc = shifts[0]
for shift in shifts:
    h = union(acc, shift)
    if not h:
        yield acc
        acc = shift
    else:
        acc = h
```

ocflib/lab/hours.py
Baisang: ^ That's probably the best way to be honest, especially with berkeley specific things like RRR week etc.

ocflib/lab/stats.py
Baisang: doesn't `frontdesk` qualify as a disaster?

setup.py
asaiacai: another option is just pull the doc as a CSV string using request or urllib, that way we don't need oauth2client or gspread

ocflib/lab/hours.py
```diff
@@ -20,10 +20,18 @@
 from datetime import time
 from datetime import timedelta
 
+import requests
+
+SHIFT_LENGTH = timedelta(hours=1)
+SHIFTS_URL = 'https://docs.google.com/spreadsheet/ccc?key=1WgczUrxqey63fmPRmdkCDMCbsfaTyCJrixiCeJj35UI&output=csv'
```
jvperrin: This key is ok to be publicly exposed, right? I'm assuming it's fine since it appears to only give read-only access, but I'd just like to make sure.

ocflib/lab/hours.py
```diff
@@ -20,10 +20,18 @@
 from datetime import time
 from datetime import timedelta
 
+import requests
+
+SHIFT_LENGTH = timedelta(hours=1)
```
jvperrin: Probably a decent assumption to make, not sure if we'd get half hour shifts at any point, but this is ok for now.

ocflib/lab/hours.py
```diff
@@ -134,6 +142,88 @@ def closed_all_day(self):
...
+        when = datetime.now()
+
+    shifts = _shift_matrix()[1:]
+    for hour in shifts:
+        if when in _shift_from_start(
+                _string_to_time(hour[0])):
+            return hour[when.weekday() + DAY_OFFSET]
+
+
 REGULAR_HOURS = defaultdict(lambda: [Hour(time(9), time(19))], {
```
jvperrin: Do you think it would be a good idea if these regular hours came from the spreadsheet somehow? Seems like it would be nice to be able to dynamically edit these and also that it's information that should be encoded in the structure of the shifts in the spreadsheet.

ocflib/lab/hours.py
```diff
@@ -134,6 +142,88 @@ def closed_all_day(self):
         return not self.hours
 
 
+def _shift_from_start(start):
```
jvperrin: This method might fit better in the `Hour` class? Not a big deal though

ocflib/lab/hours.py
```diff
@@ -134,6 +142,88 @@ def closed_all_day(self):
...
+    if shift1.close in shift2 or shift2.close in shift1:
+        return Hour(min(shift1.open, shift1.open),
+                    max(shift1.close, shift2.close))
+    else:
+        return None
+
+
+def get_hours(when=None):
+    """Combines lab shifts as much as possible and returns the final list of Hours for
+    'day'.  'day' is an int [0,6] or datetime object"""
```
jvperrin: Do you mean for `when` instead of `day`? I don't see a `day` here any more, so I'm assuming so

ocflib/lab/hours.py
asaiacai: The key is read-only.  People will need to provide some form of authentication if they want to edit the spreadsheet which requires more than just the spreadsheet key iirc

jvperrin: Ok, that's what I thought from looking at it and trying it out. Who has access to edit currently? I was thinking of adding protected ranges to the spreadsheet so that only certain parts could be edited to make it less likely to be accidently messed up.

asaiacai: Right now just me and @tahabi I can add you or we should honestly just make a new one in the OCF drive (that I don't have access to RIP).

jvperrin: I can just move it there and give you access to the drive, that seems like it makes the most sense to me.

jvperrin:
Closed by #75.

PR #75, opened 2017-08-17
Updated Big Brother implementation

tahabi:
Building off of @asaiacai's work, this tries to take into account some of the review suggestions from #74. Instead of loading hours from a hardcoded variable, hours are pulled from ocfweb but exposed in the usual `REGULAR_HOURS` format to minimize the amount of code that needs to be rewritten. Holidays work as usual because only the source of `REGULAR_HOURS` is being replaced, `Day` remains unchanged.

### DIFF
--- a/ocflib/lab/hours.py
+++ b/ocflib/lab/hours.py
@@ -13,7 +13,6 @@ Usage:
         hours=[Hour(open=9, close=21)],
     )
 """
-from collections import defaultdict
 from collections import namedtuple
 from datetime import date
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


tahabi:
the tests will fail because HOURS_URL won't return something useful for this until the ocfweb PR is merged, among other things. You can test this by swapping HOURS_URL for 'http://nyx.ocf.berkeley.edu:8934/api/hours'

ocflib/lab/hours.py
```diff
@@ -13,17 +13,46 @@
...
 from datetime import timedelta
 
+import requests
 
-class Hour(namedtuple('Hours', ['open', 'close'])):
+HOURS_URL = 'https://ocf.berkeley.edu/api/hours'
+
+
+def _pull_hours():
+    """small helper to make testing easier"""
```
chriskuehl: provides no information

and if it's only to help with testing, shouldn't it go in the tests?

ocflib/lab/hours.py
```diff
@@ -13,17 +13,46 @@
...
+    return requests.get(HOURS_URL).json()
+
+
+def _generate_regular_hours():
+    """produce hours in the manner previously exposed by the hardcoded REGULAR_HOURS variable,
+       but pull the data from ocfweb instead, where the hours come from a Google Spreadsheet."""
```
chriskuehl: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

at least follow the format:

```python
"""One-line summary.

Multi-line description following a blank line.
"""
```

i'm also not sure it's necessary to reference "the manner previously exposed" in a docstring? this comment doesn't age well. can you just describe the response format instead?

ocflib/lab/hours.py
tahabi: It's there so I can patch it to feed _generate_regular_hours() a fixed string in the tests. Is there a better way of doing this?

chriskuehl: something like this should work:

```python
@pytest.fixture
def mock_hours_response():
    with mock.patch.object(requests, 'get') as m:
        m.return_value.json.return_value = {
            'some key': 'some value',
        }
        yield
```

ocflib/lab/hours.py
tahabi: apologies. how about something like this:

    """pull hours from ocfweb and return them in the manner expected by Day()

       OCF lab hours are determined by a Google Spreadsheet, and processing
       them is handled by the ocfweb API. Originally, hours were hardcoded
       into ocflib. This is a shim for code that still expects hours to come
       from ocflib. The format is as follows, a dict of lists of Hour()s:

       REGULAR_HOURS = {
           Day.MONDAY: [Hour(time(11, 10), time(13), 'staff1'),
                        Hour(time(14, 10), time(18), 'staff2'),
                        ...],
           Day.TUESDAY: ...
           ...
       }
    """

matthew-mcallister: Capitalization and indentation. Optionally, you could frame the return format as an example:
```python
>>> _generate_regular_hours()
{
    Day.MONDAY: [Hour(time(11, 10), time(13), 'staff1'),
                 Hour(time(14, 10), time(18), 'staff2'),
                 ...],
    Day.TUESDAY: ...
    ...
}
```

kpengboy: Nitpick: the one line summary should end in a period

ocflib/lab/hours.py
```diff
@@ -13,17 +13,46 @@
...
+
+    return regular_hours
+
+
+def _parsetime(t):
+    return datetime.strptime(t, '%H:%M:%S').time()
+
+
+class Hour(namedtuple('Hours', ['open', 'close', 'staffer'])):
+    """duplicating class to ocfweb until this can be refactored"""
```
jvperrin: Add a `TODO` here so it is easier to find for future work?

ocflib/lab/hours.py
```diff
@@ -13,17 +13,46 @@
...
+def _generate_regular_hours():
+    """produce hours in the manner previously exposed by the hardcoded REGULAR_HOURS variable,
+       but pull the data from ocfweb instead, where the hours come from a Google Spreadsheet."""
+
+    regular_hours = {}
+
+    for day, hours in _pull_hours().items():
+        regular_hours[int(day)] = [Hour(open=_parsetime(hour[0]),
+                                        close=_parsetime(hour[1]),
+                                        staffer=hour[2]) for hour in hours]
```
jvperrin: Something like this might be a bit cleaner and make it clearer what is happening (plus trailing commas are nice):

    for day, hours in _pull_hours().items():
        regular_hours[int(day)] = [
            Hour(
                open=_parsetime(hour[0]),
                close=_parsetime(hour[1]),
                staffer=hour[2],
            )
            for hour in hours
        ]

ocflib/lab/hours.py
```diff
@@ -134,24 +163,28 @@ def closed_all_day(self):
...
+    (date(2017, 4, 30), date(2017, 4, 30), 'Early Closing', []),
+    (date(2017, 5, 3), date(2017, 5, 3), 'Dead week hours',
+        [Hour(time(9, 10), time(12, 30), 'test'), Hour(time(14, 10), time(19, 0), 'staff')]),
+    (date(2017, 5, 12), date(2017, 5, 12), 'Last Day Spring 2017', [Hour(time(9), time(12), 'staff')]),
     (date(2017, 5, 13), date(2017, 8, 23), 'Summer Break', []),
     (date(2017, 9, 4), date(2017, 9, 4), 'Labor Day', []),
     (date(2017, 11, 10), date(2017, 11, 10), 'Veterans Day', []),
     (date(2017, 11, 22), date(2017, 11, 26), 'Thanksgiving Break', []),
-    (date(2017, 12, 15), date(2017, 12, 15), 'Last Day Fall 2017', [Hour(time(9), time(12))]),
+    (date(2017, 12, 15), date(2017, 12, 15), 'Last Day Fall 2017', [Hour(time(9), time(12), 'staff')]),
```
jvperrin: Ideally I don't think we'd want to have specific staff members here, especially if they are placeholders that aren't actually used. Perhaps it could be made an optional attribute/argument for the Hour class instead?

ocflib/lab/hours.py
```diff
@@ -134,24 +163,28 @@ def closed_all_day(self):
...
     # start date, end date, holiday name, list of hours (date ranges are inclusive)
+    (date(2017, 3, 24), date(2017, 3, 24), 'Friday before Spring Break', [Hour(time(9), time(12), 'staff')]),
+    (date(2017, 3, 25), date(2017, 4, 2), 'Spring Break', []),
+    (date(2017, 4, 27), date(2017, 4, 27), 'MLK is closed', []),
+    (date(2017, 4, 28), date(2017, 5, 2), 'Dead week hours',
+     [Hour(time(9, 10), time(14), 'test'), Hour(time(15, 10), time(19), 'staff')]),
+    (date(2017, 4, 30), date(2017, 4, 30), 'Early Closing', []),
+    (date(2017, 5, 3), date(2017, 5, 3), 'Dead week hours',
+        [Hour(time(9, 10), time(12, 30), 'test'), Hour(time(14, 10), time(19, 0), 'staff')]),
+    (date(2017, 5, 12), date(2017, 5, 12), 'Last Day Spring 2017', [Hour(time(9), time(12), 'staff')]),
```
jvperrin: I removed these in c2f962e8c8, maybe you need to rebase?

tahabi: your rebase is in the commit history for this branch, I may have accidentally removed it in some merge conflict or something.

PR #77, opened 2017-08-23
Replace pycrypto with pycryptodome

jvperrin:
`pycryptodome` appears to be a [drop in replacement for `pycrypto`](https://www.pycryptodome.org/en/latest/src/installation.html#installation), so this should work fine, I'll see if tests pass or not. ([rt#5885](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=5885))

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,10 +28,9 @@ setup(
         'ldap3',
         'paramiko',
         'pexpect',
-        'pycrypto',
+        'pycryptodome',
         'pymysql',
-        # 4.3.3 adds a dependency on "pycryptodome" which conflicts with pycrypto
-        'pysnmp<=4.3.2',
+        'pysnmp',
         'pyyaml',
         'redis',
         'requests',


chriskuehl:
here's the usual caveat: venv-update isn't smart enough to handle removing/adding packages that provide the same files, you may need to rm any pre-existing virtualenvs

maybe replace my celery fork at the same time so that people don't have to do this twice?

jvperrin:
Yeah, good idea about updating the celery fork, I was actually just doing that, since this change also unblocks unpinning `pyasn1` in ocfweb.

I think it might be ok, since just running `make venv` in this repo worked fine to remove the old package and install the new one without removing the whole venv, but maybe ocfweb would be different?

chriskuehl:
if it works then that's cool. venv-update does not guarantee it will though, so potentially it could break /shrug

jvperrin:
Alright, well I'll keep that in mind, if something breaks with venvs I'll know which PR to blame :P

PR #79, opened 2017-09-02
reserve guest as a username

Baisang:
since we are adding shortlink /guest

see https://github.com/ocf/puppet/pull/185

### DIFF
--- a/ocflib/account/validators.py
+++ b/ocflib/account/validators.py
@@ -78,6 +78,7 @@ RESERVED_USERNAMES = frozenset((
     'gm',
     'gnats',
     'groups',
+    'guest',
     'hello',
     'help',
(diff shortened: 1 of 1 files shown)


Baisang:
wait why did this fail @jvperrin ?

jvperrin:
It looks like it didn't like the slash in the branch name for some reason?

jvperrin:
Probably because it makes a new directory per branch and having a slash in the branch name does not map well to having a directory named after it... Kinda annoying, but using a hyphen or something should fix it.

Baisang:
oh, i see what happened. yeah it looks like its html encoding the '/' because otherwise directories and stuff and yeah.

oh well should work i think

Issue #85, opened 2017-10-04
Fix `ocflib.lab.stats.list_desktops(public_only=True)`


kpengboy:
The LDAP queries are outdated since we moved the relevant setting to hiera.

@jvperrin PuppetDB can handle this, right?

matthew-mcallister:
Something like PuppetDB would be ideal, but I wouldn't hold my breath waiting on it. In the meantime, I think it would be OK to hard-code at least eruption and possibly the front desk computer in ocflib.

PR #86, opened 2017-10-04
lab.stats: Tell if desktop is public by DNS record

matthew-mcallister:
Fixes https://github.com/ocf/ocflib/issues/85. This is probably better than using PuppetDB anyways, as that would introduce more network dependencies to ocflib and we might want to have a non-public test computer that still allows non-staff logins.

### DIFF
--- a/ocflib/lab/stats.py
+++ b/ocflib/lab/stats.py
@@ -142,7 +142,7 @@ def list_desktops(public_only=False):
     if not public_only:
         filter = '(type=desktop)'
     else:
-        filter = '(&(type=desktop)(!(|(puppetVar=staff_only=true)(puppetVar=pubstaff_only=true))))'
+        filter = '(&(type=desktop)(!(|(dnsCname=frontdesk)(dnsCname=staffdesk))))'
 
     with ldap_ocf() as c:
         c.search(OCF_LDAP_HOSTS, filter, attributes=['cn'])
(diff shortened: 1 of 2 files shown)


ocflib/lab/stats.py
```diff
@@ -142,7 +142,7 @@ def list_desktops(public_only=False):
     if not public_only:
         filter = '(type=desktop)'
     else:
-        filter = '(&(type=desktop)(!(|(puppetVar=staff_only=true)(puppetVar=pubstaff_only=true))))'
+        filter = '(&(type=desktop)(!(|(dnsCname=frontdesk)(dnsCname=staffdesk))))'
```
chriskuehl: maybe these should be A records instead? i could see us potentially having multiple staff desktops (e.g. we at one point talked about having one in the server room as a testing record)

matthew-mcallister: Oh, you mean like `staffdesk.o.b.e` would have multiple A records? Yeah, that would be more flexible; I was expecting us to add a new CNAME here for every non-public desktop we add.

matthew-mcallister:
I would have but I don't know if that will break big brother or whatever it's called.

jvperrin:
Big brother isn't really a thing yet, it's still being worked on, so it should be fine if you'd like to switch to using A records instead of CNAMEs.

PR #87, opened 2017-10-08
Bump pre-commit hook versions

jvperrin:
Most of the changes are automatic, except for `tests-manual/misc/shell`, which I pretty much rewrote from scratch to remove the hackier parts of it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-exclude: (^vendor/|^\.activate\.sh$)
+exclude: ^vendor/
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks.git
     sha: v0.9.5
@@ -36,4 +36,3 @@ repos:
     sha: v1.1.1
     hooks:
     -   id: remove-tabs
-        exclude: ^(Makefile|debian/rules)$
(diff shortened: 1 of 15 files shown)


.pre-commit-config.yaml
```diff
@@ -1,41 +1,39 @@
+exclude: (^vendor/|^\.activate\.sh$)
```
chriskuehl: don't need to exclude symlinks like .activate.sh with new versions of pre-commit, most hooks now only target "file" type (rather than in previous versions where they just targeted all files, including symlinks and submodules)

slightly more details at https://github.com/chriskuehl/identify#usage

.pre-commit-config.yaml
```diff
@@ -1,41 +1,39 @@
...
     -   id: pyupgrade
-        language_version: python3
         args: ['--py3-plus']
 -   repo: https://github.com/Lucas-C/pre-commit-hooks.git
-    sha: v1.0.1
+    sha: v1.1.1
     hooks:
     -   id: remove-tabs
-        exclude: ^(Makefile|debian/rules|\.activate\.sh)$
+        exclude: ^(Makefile|debian/rules)$
```
chriskuehl: i think this is the default: https://github.com/Lucas-C/pre-commit-hooks/blob/3e5968a57a3720a490ff23b8c1a15d874bbd97fc/.pre-commit-hooks.yaml#L27

.pre-commit-config.yaml
jvperrin: Ooh, that's pretty nice

.pre-commit-config.yaml
jvperrin: Ah yeah, much better

PR #88, opened 2017-10-09
Catch duplicate account request errors

chriskuehl:


ocflib/account/submission.py
```diff
@@ -197,14 +198,21 @@ def validate_then_create_account(request):
...
 
-                with get_session() as session:
-                    session.add(stored_request)  # TODO: error handling
-                    session.commit()
+                try:
+                    with get_session() as session:
+                        session.add(stored_request)  # TODO: error handling
+                        session.commit()
+                except sqlalchemy.exc.IntegrityError:
+                    # If there's an integrity constaint, it's okay -- the
```
matthew-mcallister: "constaint"

jvperrin: Pretty sure I would have just read over that without noticing

chriskuehl: typing is hard

matthew-mcallister: @jvperrin Yeah, I guess could have ignored that since it's only a comment and not published anywhere.

PR #99, opened 2017-11-06
Close on actual Veterans Day as well

Baisang:
no one is going to be here so we made 11-11 OCF datathon + LAN party at night


Baisang:
Gotta pad my github stats

also in general usually just want to have someone verify that we all agreed on this if possible, so thats why i make PRs for hours

PR #102, opened 2017-11-16
Add recommender.py

tliu22:
Related to https://github.com/ocf/ocfweb/pull/345

### DIFF
--- a/ocflib/account/recommender.py
+++ b/ocflib/account/recommender.py
@@ -1,0 +1,24 @@
+from ocflib.account.creation import validate_username
+
+from random import randint
+
+def recommend(first_name, last_name, n):
+    first_name = first_name.lower()
+    last_name = last_name.lower()
+
+    recommendations = []
+    attempts = 0
+    while len(recommendations) < n and attempts < 20:
+        first_rand = randint(1, len(first_name))
+        last_rand = randint(1, len(last_name))
+        first_generated = first_name[:first_rand]
+        last_generated = last_name[:last_rand]
+        recommendation = first_generated+last_generated
+        if recommendation not in recommendations:
+            try:
+                validate_username(recommendation, first_name+last_name)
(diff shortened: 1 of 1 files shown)


ocflib/account/recommender.py
```diff
@@ -0,0 +1,24 @@
+from ocflib.account.creation import validate_username
+
+from random import randint
+
+def recommend(first_name, last_name, n):
```
matthew-mcallister: This needs a doc comment.

ocflib/account/recommender.py
```diff
@@ -0,0 +1,24 @@
...
+    last_name = last_name.lower()
+
+    recommendations = []
+    attempts = 0
+    while len(recommendations) < n and attempts < 20:
+        first_rand = randint(1, len(first_name))
+        last_rand = randint(1, len(last_name))
+        first_generated = first_name[:first_rand]
+        last_generated = last_name[:last_rand]
+        recommendation = first_generated+last_generated
```
matthew-mcallister: Do you think this is really a good recommendation system? I think conventional choices should be listed first, e.g. first name + last name, first initial + last name, first name + last initial, first-middle-last initials. I would only consider using truncated portions of names if all the preceding combinations are taken.

abizer:
I agree with Matt, this should live entirely in ocfweb. 

Also, for future reference, do `make venv` and `make install-hooks` and pre-commit will catch the formatting errors and warn you before you commit it.

ocflib/account/recommender.py
tliu22: I originally had that the conventional choices listed first (the script someone else wrote did), but since we had a refresh/more button for more recommendations, we'd have to keep state somehow, so for testing purposes I just drafted this up. But yea, I'll work up a better recommendation script.

PR #103, opened 2017-11-21
consolidate get_connection across ocflib

abizer:


ocflib/misc/db.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def get_connection(user,
+                   password,
+                   db,
+                   host='mysql.ocf.berkeley.edu',
+                   cursorclass=pymysql.cursors.DictCursor,
+                   charset='utf8mb4',
+                   autocommit=True,
+                   **kwargs):
```
kpengboy: Might want to put a short doc comment here

abizer: done

ocflib/misc/db.py
```diff
@@ -0,0 +1,25 @@
+import pymysql
```
jvperrin: I feel like this file fits a bit better in `ocflib/infra`, since the other files in that module are there for connecting to different OCF services, like [LDAP](https://github.com/ocf/ocflib/blob/master/ocflib/infra/ldap.py), [kerberos](https://github.com/ocf/ocflib/blob/master/ocflib/infra/kerberos.py), etc. Having the database connection module in there makes a bit more sense to me than having it in misc.

ocflib/misc/db.py
```diff
@@ -0,0 +1,25 @@
...
+    in some ocflib modules that may be better suited for particular tasks."""
+
+    return pymysql.connect(
+        user=user,
+        password=password,
+        db=db,
+        host=host,
+        cursorclass=cursorclass,
+        charset=charset,
+        autocommit=autocommit
```
jvperrin: Aren't you throwing away the `kwargs` here since you don't use them anywhere?

jvperrin: Also, I'm not sure why `**dict({'autocommit': True}, **kwargs)` was used in the existing functions instead of using something simpler like `autocommit=autocommit, **kwargs`, so I'd be hesitant to change anything until that is figured out just because I don't want to break anything that uses these functions.

abizer: The only time it's used is in `ocfweb/account/vhost-mail.py` and the syntax is compatible with this format. I guess I am throwing away the kwargs, but we don't use them anywhere anyways AFAIK that isn't already captured here. I'll put them back in just in case.

ocflib/misc/db.py
abizer: yeah you're right that was pretty stupid of me

ocflib/misc/db.py
jvperrin: I think having them makes sense, since then any other arguments get passed along too. Good point though on the syntax being compatible with the existing calls, I think that's probably fine then.

ocflib/misc/db.py
jvperrin: Nah, this isn't a bad place to put it, I just think there's a slightly better place :)

chriskuehl:
This is fine, but eventually we should take the opportunity to fix this API. Currently it's really hard to write correct code which closes connections as you'd expect; a better API would be something like this:

```python
@contextlib.contextmanager
def get_connection(...):
    with contextlib.closing(pymysql.connect(...)) as conn:
        yield conn
```

Then usage is like this:

```python
with get_connection() as conn:
    with conn as cursor:
        cursor.execute('...')
```

jvperrin:
Oh lol, I thought we closed them again, but it makes some sense we don't since we just return a connection that is never actually closed anywhere. That does sound like a good idea to have something to close the connections when they are done.

abizer:
I'll put it on my todo list to start upgrading the API across the codebase.

jvperrin:
Alright, I've opened an issue in #108, since I think that's a bit of a separate issue, but definitely one that should be addressed sometime.

PR #105, opened 2017-11-21
Fix flake8 except errors

kpengboy:


ocflib/ucb/groups.py
```diff
@@ -154,10 +154,7 @@ def _get_osl(query, service, parser):
 def _parse_osl(root, parser):
     """Assemble Python dictionaries of groups from XML document"""
     if root.findtext('Succeeded') == 'false':
-        try:
-            error_reason = root.findtext('Reason')
-        except:
-            error_reason = 'unknown reason'
+        error_reason = root.findtext('Reason') or 'unknown_reason'
```
jvperrin: This won't raise an error at all?

kpengboy: Would appreciate an example, I could not find one

kpengboy: Actually I could've just done `error_reason = root.findtext('Reason', default='unknown reason')`

jvperrin: I'm guessing it would be a [`ParseError`](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.ParseError), but I'm not really sure, I haven't ever seen this output `unknown reason`, so I'm not sure when it even fails.

jvperrin: Yeah, that makes sense, I think using the default would be cleaner for this.

jvperrin:
Well, looks like in the meantime, @matthew-mcallister changed this in 392cc0fe73c15be09f12a41a8e4a522c03d415aa anyway

ocflib/ucb/groups.py
matthew-mcallister: I agree, feel free to overwrite my mistaken commit. For the record, it doesn't seem like this can raise an exception according to the documentation or the [source code](https://github.com/python/cpython/blob/3.6/Lib/xml/etree/ElementPath.py#L309).

PR #107, opened 2017-11-22
ocfstats sql move from labstats

jvperrin:
This moves the bulk of the `ocfstats` table definition from [labstats/bin/init.py](https://github.com/ocf/labstats/blob/master/labstats/bin/init.py) to be in ocflib, like the rest of the SQL tables.


ocflib/lab/ocfstats.sql
```diff
@@ -0,0 +1,103 @@
...
+    );
+
+CREATE VIEW `users_in_lab` AS
+    SELECT `user`, `host`, `start` FROM `session` WHERE `end` IS NULL;
+
+CREATE VIEW `users_in_lab_count_public` AS
+    SELECT COUNT(DISTINCT `user`) AS `count` FROM `users_in_lab`;
+
+--- This relies on the semester_start function, which is defined in ocfprinting.sql
+CREATE VIEW `unique_users_in_lab_count_public` AS
```
kpengboy: Name of this view sounds confusingly similar to `users_in_lab_count_public`, even though they count very different things...

ocflib/lab/ocfstats.sql
```diff
@@ -0,0 +1,103 @@
+---
+--- Documentation of the ocfstats database schema
+---
+
```
kpengboy: Also do we want to add `DROP TABLE IF EXISTS` and `DROP VIEW IF EXISTS` statements?

ocflib/lab/ocfstats.sql
abizer: This was another semi-temporary table created for testing something @tanx16 was working on for his family project, he was planning on doing this same thing eventually

ocflib/lab/ocfstats.sql
jvperrin: I'll add the `DROP VIEW IF EXISTS` statements, but I'm pretty hesitant to add any `DROP TABLE IF EXISTS` statements, because I'd rather not lose the existing tables data if someone runs it unknowingly. The views are fine to remove and re-add though, so those would be fine. This also mirrors the current behavior in [ocfmail.sql](https://github.com/ocf/ocflib/blob/master/ocflib/vhost/ocfmail.sql) and [ocfprinting.sql](https://github.com/ocf/ocflib/blob/master/ocflib/printing/ocfprinting.sql), so that seems like a good idea.

jvperrin:
@abizer Oh ok, I didn't realize that was the case, I figured moving it was just a good idea to try and move closer to removing labstats.

PR #110, opened 2017-12-01
Added periods to validate exceptions

tliu22:

PR #111, opened 2017-12-02
Added larger list from www.bannedwordlist.com/lists/swearWords.csv

tliu22:
Added larger list from www.bannedwordlist.com/lists/swearWords.csv, especially in light of quick-validation and recommendation system.

### DIFF
--- a/ocflib/account/creation.py
+++ b/ocflib/account/creation.py
@@ -29,9 +29,9 @@ from ocflib.printing.quota import SEMESTERLY_QUOTA
 
 
 _KNOWN_UID = 50500
-BAD_WORDS = frozenset(('anal', 'anus', 'arse', 'ass', 'ballsack', 'balls', 'bastard', 'bitch', 'biatch', 'bloody', 'blowjob', 'blow job', 'bollock', 'bollok', 'boner', 'boob', 'bugger', 'bum', 'butt', 'buttplug', 'clitoris', 'cock', 'coon',
-'crap', 'cunt', 'damn', 'dick', 'dildo', 'dyke', 'fag', 'feck', 'fellate', 'fellatio', 'felching', 'fuck', 'fudgepacker', 'flange', 'Goddamn', 'hell', 'homo', 'jerk', 'jizz', 'knobend', 'knob end', 'labia', 'lmao', 'lmfao', 'muff',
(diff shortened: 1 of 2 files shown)


sahilhasan:
I would personally say we should remove lmao, lmfao and wtf. None of those are 'bad' enough to warrant not allowing imo

ocflib/account/creation.py
```diff
@@ -29,7 +29,12 @@
 
 
 _KNOWN_UID = 50500
-BAD_WORDS = frozenset(('fuck', 'shit', 'cunt', 'bitch', 'dick'))
+BAD_WORDS = frozenset(('anal', 'anus', 'arse', 'ass', 'ballsack', 'balls', 'bastard', 'bitch', 'biatch', 'bloody',
+'blowjob', 'bollock', 'bollok', 'boner', 'boob', 'bugger', 'bum', 'butt', 'buttplug', 'clitoris', 'cock', 'coon',
+'crap', 'cunt', 'damn', 'dick', 'dildo', 'dyke', 'fag', 'feck', 'fellate', 'fellatio', 'felching', 'fuck',
+'fudgepacker', 'flange', 'Goddamn', 'hell', 'homo', 'jerk', 'jizz', 'kike', 'knobend', 'labia', 'lmao', 'lmfao', 'muff',
```
matthew-mcallister: Why have "Goddamn" if "damn" already appears? Also that "G" shouldn't be capitalized.

ocflib/account/creation.py
```diff
@@ -327,13 +332,13 @@ def validate_username(username, realname):
 
     if similarity_heuristic(realname, username) > SIMILARITY_THRESHOLD:
         raise ValidationWarning(
-            'Username {} not based on real name {}'.format(username, realname))
+            'Username {} not based on real name {}.'.format(username, realname))
```
matthew-mcallister: Isn't this https://github.com/ocf/ocflib/pull/110? These changes shouldn't go here I don't think.

matthew-mcallister:
Alright, this list is not bad. I think some duplicate words need removing (see my comment for an example). Also, the indentation should be fixed to look like
```python
xyz = frozenset({
    'foo', 'bar', 'baz', ...
})
```

I agree with @sahilhasan's comments. In particular, `wtf` are valid initials and `lmao` could be based on a real name (i.e. last name Mao).

tliu22:
Oh, I might've missed some duplicate words. Are there any more?

And I thought 'wtf' and 'lmao' were already removed?

sahilhasan:
@tliu22 RIP I looked at the CSV (and I remembered from the previous list). Thats my fault. Sry

matthew-mcallister:
@tliu22 "balls" and "ballsack"

kpengboy:
Late to the game, but we should also expand the test cases to cover these new bad words.

kpengboy:
And we need to get the Scunthorpe project done soon...

PR #112, opened 2017-12-15
remove ocflib.lab.printing

asaiacai:


abizer:
This is still in use; you should migrate it in ocfweb before merging this:

```
$ ag ocflib.lab.printing
ocfweb/ocfweb/stats/summary.py
7:from ocflib.lab.printing import get_maintkit
8:from ocflib.lab.printing import get_toner
9:from ocflib.lab.printing import PRINTERS
```

PR #113, opened 2017-12-15
move mirror stats to ocflib

asaiacai:


ocflib/lab/stats.py
```diff
@@ -39,6 +39,24 @@ def duration(self):
 UserTime = namedtuple('UserTime', ['user', 'time'])
 
 
+def _humanize(n):
```
jvperrin: I'd just name this `humanize_bytes` or something (to clarify and to remove the underscore). Then it can be used in `upload-to-box` too so it's not repeated in both places!

ocflib/lab/stats.py
```diff
@@ -39,6 +39,24 @@ def duration(self):
...
+    # adapted from jvperrin/upload-to-box
```
jvperrin: I'd leave the [comment from the `upload-to-box` script](https://github.com/ocf/puppet/blob/beb9090d1bef0f428a0a32996b0186c91bb74013/modules/ocf_backups/files/upload-to-box#L282-L284) here, since I think it's more useful and says what the function actually does.

PR #115, opened 2018-02-08
message brokering with RabbitMQ

asaiacai:
Eventual goal is to have RabbitMQ run somewhere in our infra and use it for sending desktop notifications from the printer.  Trying to make it extensible for other message brokering use cases.


asaiacai:
Dropping this since it sounds like people were in favor of using Redis pubsub

PR #117, opened 2018-02-13
databasing holidays

asaiacai:
Eventual goal is to have web front-end to add holidays / temp hour changes be databased instead of having to make PR's everytime

### DIFF
--- a/ocflib/lab/hours.py
+++ b/ocflib/lab/hours.py
@@ -21,8 +21,10 @@ from datetime import timedelta
 
 import requests
 
-HOURS_URL = 'https://www.ocf.berkeley.edu/api/hours'
+from ocflib.infra.db import get_connection
+from ocflib.printing.quota import _namedtuple_to_query
 
+HOURS_URL = 'https://www.ocf.berkeley.edu/api/hours'
 
 def _generate_regular_hours():
     """pull hours from ocfweb and return them in the manner expected by Day().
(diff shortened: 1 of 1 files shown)


ocflib/lab/hours.py
```diff
@@ -60,6 +62,35 @@ def _generate_regular_hours():
...
+    """Get a list of holidays from a database"""
+    raw = _raw_holidays(limit)
+    holidays = []
+    for day in raw:
+        d = (day['start_date'],
+            day['end_date'],
+            day['holiday'],
+            (datetime.min + day['start_time']).time(),
+            (datetime.min + day['end_time']).time()
+            )
```
dkess: A [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) would be better here.

ocflib/lab/hours.py
```diff
@@ -60,6 +62,35 @@ def _generate_regular_hours():
 def _parsetime(t):
     return datetime.strptime(t, '%H:%M:%S').time()
 
+def _get_holiday_connection():
+    return get_connection(user='anonymous',
+            password=None,
+            db='holiday')
```
chriskuehl: bizarre indentation

ocflib/lab/hours.py
```diff
@@ -60,6 +62,35 @@ def _generate_regular_hours():
...
+def add_holiday(holiday):
+    """Adds a holiday to the database"""
+    
+    with _get_holiday_connection() as c:
+        c.execute(*_namedtuple_to_query('INSERT INTO holiday ({}) VALUES({})', holiday))
+
+def _raw_holidays(limit):
+    with _get_holiday_connection() as c:
+        c.execute("SELECT * FROM holiday")
+        return c.fetchmany(size=limit)
```
chriskuehl: this should return namedtuples rather than raw row dicts

chriskuehl:
can we move this code to ocfweb if we're going to do this?

it would be better for ocfweb to provide an API (and talk to mysql itself) then to have every single thing that needs hours to talk to mysql

we should strive to have one thing provide an interface over top of a backing datastore rather than everything reaching into the datastore.

ocflib/lab/hours.py
```diff
@@ -60,6 +62,35 @@ def _generate_regular_hours():
...
+
+def add_holiday(holiday):
+    """Adds a holiday to the database"""
+    
+    with _get_holiday_connection() as c:
+        c.execute(*_namedtuple_to_query('INSERT INTO holiday ({}) VALUES({})', holiday))
+
+def _raw_holidays(limit):
+    with _get_holiday_connection() as c:
+        c.execute("SELECT * FROM holiday")
```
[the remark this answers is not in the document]
kpengboy: I concur, it turns out that `fetchmany` can actually be slower than `fetchall` with a `LIMIT` because fetchmany will cache many additional queries beyond the amount you want. Granted, it probably won't be as big of a problem here compared to fetching from the desktop logins table, but I'd say it's better practice to make the change.

asaiacai:
I'll move this to ocfweb

ocflib/lab/hours.py
```diff
@@ -60,6 +62,35 @@ def _generate_regular_hours():
...
+def _get_holiday_connection():
+    return get_connection(user='anonymous',
+            password=None,
+            db='holiday')
+
+def add_holiday(holiday):
+    """Adds a holiday to the database"""
+    
+    with _get_holiday_connection() as c:
+        c.execute(*_namedtuple_to_query('INSERT INTO holiday ({}) VALUES({})', holiday))
```
chriskuehl: imo this is more readable if you just pass in the args by hand rather than using a utility function

PR #118, opened 2018-02-14
move namedtuple_to_query

asaiacai:
name says it all. refactoring to use in databasing temp hour changes in ocfweb


chriskuehl:
do we really need to generalize this? what's the new situation you want to use this in?

generally speaking i think this is a net-negative for readability and should be avoided unless it's really needed (and yes, i wrote the code to begin with). unless we're really doing this over and over and over, it would imo be better to just construct/destruct the namedtuple by hand.

PR #123, opened 2018-05-02
happy hour spring 2018

abizer:
since we're not changing the quota, I don't believe a BoD vote
is necessary for this change. Presumably "Happy Hour" has some
pre-existing standing such that we don't need BoD approval
to enable it during the expected times it would be enabled.

At least, this is mine and @sahilhasan's understanding.


kpengboy:
Hmm...it seems that this has been the de facto practice for a few semesters now, but it doesn't seem that this has been explicitly allowed in the past. There was also at least one instance (2014-12-01) where BoD instituted happy hours. I think it would be better for BoD to clarify the rules here - while the Cabinet can already amend the quota _in between_ semesters, perhaps we should allow the Cabinet to amend the printing quota before the first BoD meeting of a semester and after the last BoD meeting?

Or we could explicitly codify the concept of happy hours and require BoD's approval for any other changes to the quota at any other time.

PR #126, opened 2018-07-07
approve: handle usernames not in session

### LINKED ISSUE #81: create throws an uncaught exception when you reject a username that doesn't exist
```
A problem was encountered and reported via ocflib:

An exception occured in create:

Traceback (most recent call last):
  File "/opt/create/venv/lib/python3.5/site-packages/sqlalchemy/orm/session.py", line 1710, in delete
    state = attributes.instance_state(instance)
AttributeError: 'NoneType' object has no attribute '_sa_instance_state'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/create/venv/lib/python3.5/site-packages/celery/app/trace.py", line 367, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/opt/create/venv/lib/python3.5/site-packages/celery/app/trace.py", line 622, in __protected_call__
    return self.run(*args, **kwargs)
  File "/opt/create/venv/lib/python3.5/site-packages/ocflib/account/submission.py", line 304, in reject_request
    stored_request = get_remove_row_by_user_name(user_name)
  File "/opt/create/venv/lib/python3.5/site-packages/ocflib/account/submission.py", line 292, in get_remove_row_by_user_name
    session.delete(request_row)
  File "/opt/create/venv/lib/python3.5/site-packages/sqlalchemy/orm/session.py", line 1712, in delete
    raise exc.UnmappedInstanceError(instance)
sqlalchemy.orm.exc.UnmappedInstanceError: Class 'builtins.NoneType' is not mapped


Task Details:
  * task_id: 9853248b-ee73-49c2-902c-0199d59033ba

Try `journalctl -u ocf-create` for more details.

====
Hostname: 41424c6fc0e1
Callstack:
    at /opt/create/venv/lib/python3.5/site-packages/ocflib/misc/mail.py:94 (send_problem_report)
        by /opt/create/create/tasks.py:88 (failure_handler)
        by /opt/create/venv/lib/python3.5/site-packages/celery/app/trace.py:205 (handle_failure)
        by /opt/create/venv/lib/python3.5/site-packages/celery/app/trace.py:157 (handle_error_state)
        by /opt/create/venv/lib/python3.5/site-packages/celery/app/trace.py:323 (on_error)
        by /opt/create/venv/lib/python3.5/site-packages/celery/app/trace.py:381 (trace_task)
        by /opt/create/venv/lib/python3.5/site-packages/celery/app/trace.py:530 (_fast_trace_task)
        by /opt/create/venv/lib/python3.5/site-packages/billiard/pool.py:359 (workloop)
        by /opt/create/venv/lib/python3.5/site-packages/billiard/pool.py:290 (__call__)
        by /opt/create/venv/lib/python3.5/site-packages/billiard/process.py:112 (run)
        by /opt/create/venv/lib/python3.5/site-packages/billiard/process.py:306 (_bootstrap)
        by /opt/create/venv/lib/python3.5/site-packages/billiard/popen_fork.py:79 (_launch)
        by /opt/create/venv/lib/python3.5/site-packages/billiard/popen_fork.py:24 (__init__)
        by /opt/create/venv/lib/python3.5/site-packages/billiard/context.py:333 (_Popen)
        by /opt/create/venv/lib/python3.5/site-packages/billiard/process.py:122 (start)
        by /opt/create/venv/lib/python3.5/site-packages/billiard/pool.py:1117 (_create_worker_process)
        by /opt/create/venv/lib/python3.5/site-packages/celery/concurrency/asynpool.py:439 (_create_worker_process)
        by /opt/create/venv/lib/python3.5/site-packages/billiard/pool.py:1008 (__init__)
        by /opt/create/venv/lib/python3.5/site-packages/celery/concurrency/asynpool.py:422 (__init__)
        by /opt/create/venv/lib/python3.5/site-packages/celery/concurrency/prefork.py:112 (on_start)
        by /opt/create/venv/lib/python3.5/site-packages/celery/concurrency/base.py:131 (start)
        by /opt/create/venv/lib/python3.5/site-packages/celery/bootsteps.py:370 (start)
        by /opt/create/venv/lib/python3.5/site-packages/celery/bootsteps.py:119 (start)
        by /opt/create/venv/lib/python3.5/site-packages/celery/worker/worker.py:203 (start)
        by /opt/create/venv/lib/python3.5/site-packages/celery/bin/worker.py:256 (run)
        by /opt/create/venv/lib/python3.5/site-packages/celery/bin/base.py:244 (__call__)
        by /opt/create/venv/lib/python3.5/site-packages/celery/bin/worker.py:221 (run_from_argv)
        by /opt/create/venv/lib/python3.5/site-packages/celery/bin/celery.py:412 (execute)
        by /opt/create/venv/lib/python3.5/site-packages/celery/bin/celery.py:480 (handle_argv)
        by /opt/create/venv/lib/python3.5/site-packages/celery/bin/base.py:281 (execute_from_commandline)
        by /opt/create/venv/lib/python3.5/site-packages/celery/bin/celery.py:488 (execute_from_commandline)
        by /opt/create/venv/lib/python3.5/site-packages/celery/bin/celery.py:326 (main)
        by /opt/create/venv/lib/python3.5/site-packages/celery/__main__.py:14 (main)
        by /opt/create/venv/bin/celery:11 (<module>)
```

kkuehlz:
both of the added tests will fail prior to this patch.

fixes #81

### DIFF
--- a/ocflib/account/submission.py
+++ b/ocflib/account/submission.py
@@ -297,19 +297,25 @@ def get_tasks(celery_app, credentials=None):
             request_row = session.query(StoredNewAccountRequest).filter(
                 StoredNewAccountRequest.user_name == user_name
             ).first()
-            session.delete(request_row)
-            session.commit()
+            if request_row:
+                session.delete(request_row)
+                session.commit()
             return request_row
 
     @celery_app.task
     def approve_request(user_name):
-        request = get_remove_row_by_user_name(user_name).to_request()
+        stored_request = get_remove_row_by_user_name(user_name)
+        if not stored_request:
+            raise ValueError('No pending request for user')
+        request = stored_request.to_request()
         create_account.delay(request)
         dispatch_event('ocflib.account_approved', request=request.to_dict())
 
     @celery_app.task
     def reject_request(user_name):
         stored_request = get_remove_row_by_user_name(user_name)
+        if not stored_request:
+            raise ValueError('No pending request for user')
         request = stored_request.to_request()
         send_rejected_mail(request, stored_request.reason)
         dispatch_event('ocflib.account_rejected', request=request.to_dict())
--- a/tests/account/submission_test.py
+++ b/tests/account/submission_test.py
@@ -123,6 +123,17 @@ def test_approve_request(celery_app, fake_new_account_request, session_with_requ
     ]
 
 
+def test_approve_request_invalid(celery_app, fake_new_account_request, session_with_requests, tasks):
+    assert len(session_with_requests.query(StoredNewAccountRequest).all()) == 2
+    with pytest.raises(ValueError):
+        tasks.approve_request('keur')
+
+    # invalid request, nothing changed
+    assert len(session_with_requests.query(StoredNewAccountRequest).all()) == 2
+    tasks.create_account.delay.assert_not_called()
+    assert celery_app._sent_messages == []
+
+
 @mock.patch('ocflib.account.submission.send_rejected_mail')
 def test_reject_request(send_rejected_mail, celery_app, fake_new_account_request, session_with_requests, tasks):
     tasks.reject_request(fake_new_account_request.user_name)
@@ -137,6 +148,19 @@ def test_reject_request(send_rejected_mail, celery_app, fake_new_account_request
 
     # TODO: assert real reason
     send_rejected_mail.assert_called_once_with(request, mock.ANY)
+
+
+@mock.patch('ocflib.account.submission.send_rejected_mail')
+def test_reject_request_invalid(send_rejected_mail, celery_app, fake_new_account_request,
+                                session_with_requests, tasks):
+    assert len(session_with_requests.query(StoredNewAccountRequest).all()) == 2
+    with pytest.raises(ValueError):
+        tasks.reject_request('keur')
+
+    # invalid request, nothing changed
+    assert len(session_with_requests.query(StoredNewAccountRequest).all()) == 2
+    assert celery_app._sent_messages == []
+    send_rejected_mail.assert_not_called()
 
 
 def test_get_pending_requests(session_with_requests, tasks, fake_new_account_request):


jvperrin (dismissed):
Nice, looks good to me, I like how you handled this.

abizer (dismissed):

emmatyping (dismissed):

ocflib/account/submission.py
```diff
@@ -297,19 +297,25 @@ def get_remove_row_by_user_name(user_name):
...
+                session.delete(request_row)
+                session.commit()
             return request_row
 
     @celery_app.task
     def approve_request(user_name):
-        request = get_remove_row_by_user_name(user_name).to_request()
+        stored_request = get_remove_row_by_user_name(user_name)
+        if not stored_request:
+            return
```
chriskuehl: should this return an error so that the caller knows about this? i think it should

ocflib/account/submission.py
```diff
@@ -297,19 +297,25 @@ def get_remove_row_by_user_name(user_name):
...
+        request = stored_request.to_request()
         create_account.delay(request)
         dispatch_event('ocflib.account_approved', request=request.to_dict())
 
     @celery_app.task
     def reject_request(user_name):
         stored_request = get_remove_row_by_user_name(user_name)
+        if not stored_request:
+            return
```
chriskuehl: ditto

ocflib/account/submission.py
chriskuehl: looking at this from an interface perspective and ignoring celery and all the other details, i'd expect that if i call a function `approve_request(username)` and there's no request to be approved with that username, that it raises an exception of some kind, since it wasn't able to approve any request.

i actually think that raising an exception here is probably the right approach, but i get that it's a bit inconvenient because of the way we do error reporting from the create workers (i assume the point of this branch is to avoid the ocflib error emails?).

turns out we actually already exempt `ValueError`s from the reporting "middleware", since they usually indicate caller error: https://github.com/ocf/create/blob/f937a2b88984a2b2ffd282421dbf4e8e429c9a4a/create/tasks.py#L60-L69

maybe we should just raise a `ValueError` here? it won't generate an ocflib report, and callers (ocfweb or ircbot or whatever) would be able to catch and handle the exception if they want to. if they *don't* catch the exception, it would (ideally) trigger a crash in ocfweb/ircbot/etc, which is appropriate since this is sort of a caller error.

(i don't think it will actually trigger an error in ircbot, since i don't think it waits to see the result of the call /shrug)

does that seem reasonable? it's possible that my expectations for how this function should behave are inconsistent with other peoples'.

jvperrin: I do think raising an exception makes more sense, given that is really an error from the point of view of the function or any caller. It seems to me that relying on `ValueError`s to just not be reported is counter-intuitive though since I don't think anyone would expect that they are just ignored but all other errors are reported.

Maybe longer-term we should have a custom error type that isn't reported for this kind of thing so that errors can be raised that are expected to be handled by callers instead of sending ocflib reports?

kkuehler: Yeah, I was thinking about an exception when I first made this PR but it felt weird that, in order to stop an exception, we raise one somewhere else. After fleshing out the options it does make the most sense though.

kkuehler: This PR does feel kinda useless for the time being. The caller is gonna catch a `ValueError` as opposed to an `AttributeError`. That said, this if do go the custom exception route in the future it will make a lot more sense...

kkuehler:
@chriskuehl @jvperrin fixed and force pushed.

the first build failed i think because [jenkins builds in tmpfs](https://github.com/ocf/puppet/commit/b30b40717c5f81586d4f8c089013b30a9b9a0111) and supernova `/tmp` was full.

ocflib/account/submission.py
```diff
@@ -297,19 +297,25 @@ def get_remove_row_by_user_name(user_name):
...
+                session.delete(request_row)
+                session.commit()
             return request_row
 
     @celery_app.task
     def approve_request(user_name):
-        request = get_remove_row_by_user_name(user_name).to_request()
+        stored_request = get_remove_row_by_user_name(user_name)
+        if not stored_request:
+            raise ValueError('User not in session')
```
jvperrin: I'd say something like `"No pending request for user"` or something, I don't think the current message is really descriptive of what the problem actually is (what session?).

ocflib/account/submission.py
```diff
@@ -297,19 +297,25 @@ def get_remove_row_by_user_name(user_name):
...
+        request = stored_request.to_request()
         create_account.delay(request)
         dispatch_event('ocflib.account_approved', request=request.to_dict())
 
     @celery_app.task
     def reject_request(user_name):
         stored_request = get_remove_row_by_user_name(user_name)
+        if not stored_request:
+            raise ValueError('User not in session')
```
jvperrin: (same here)

jvperrin (commented):
Looks good to me besides making the error message a bit more descriptive, because right now I don't think it makes it that clear what is going on.

PR #128, opened 2018-07-12
Add users_in_group function to account/search

dkess:
Function that returns the list of users in an LDAP group.

### DIFF
--- a/ocflib/account/search.py
+++ b/ocflib/account/search.py
@@ -2,6 +2,7 @@
 import ldap3
 
 import ocflib.infra.ldap as ldap
+from ocflib.infra.ldap import OCF_LDAP_GROUPS
 from ocflib.infra.ldap import OCF_LDAP_PEOPLE
 from ocflib.infra.ldap import UCB_LDAP_PEOPLE
 
@@ -70,3 +71,16 @@ def user_is_group(username):
     """Returns whether username is an OCF group account."""
     attrs = user_attrs(username)
     return 'callinkOid' in attrs
+
+
+def users_in_group(groupname):
+    """Returns the list of users in an LDAP group."""
+
+    with ldap.ldap_ocf() as c:
+        c.search(
+            OCF_LDAP_GROUPS,
+            '(cn={})'.format(groupname),
+            attributes=('memberUid',),
+            search_scope=ldap3.LEVEL,
+        )
+        return c.response[0]['attributes']['memberUid']
--- a/tests/account/search_test.py
+++ b/tests/account/search_test.py
@@ -9,6 +9,7 @@ from ocflib.account.search import user_is_sorried
 from ocflib.account.search import users_by_callink_oid
 from ocflib.account.search import users_by_calnet_uid
 from ocflib.account.search import users_by_filter
+from ocflib.account.search import users_in_group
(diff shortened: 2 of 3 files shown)


ocflib/account/search.py
```diff
@@ -70,3 +71,16 @@ def user_is_group(username):
     """Returns whether username is an OCF group account."""
     attrs = user_attrs(username)
     return 'callinkOid' in attrs
+
+
+def users_in_group(groupname):
+    """Returns the list of users in an LDAP group."""
+
```
chriskuehl: why empty line?

https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

abizer:
Have you looked into adapting ocflib.account.util.list_staff? we should probably make that general purpose

tests/account/search_test.py
```diff
@@ -101,3 +102,13 @@ def test_user_is_group(self, user, exists):
...
+
+
+@pytest.mark.parametrize('user,group,is_in', [
+    ('ckuehl', 'ocfroot', True),
+    ('gstaff', 'ocfstaff', True),
+    ('guser', 'ocfstaff', False),
+    ('testopstaff', 'opstaff', True),
+])
+def test_user_is_in_group(user, group, is_in):
+    assert (user in users_in_group(group)) == is_in
```
chriskuehl: use `is` for boolean comparison (there's a flake8 rule for this but it's not smart enough to know that `is_in` is always a boolean here)

tests/account/search_test.py
```diff
@@ -101,3 +102,13 @@ def test_user_is_group(self, user, exists):
     def test_notexistent_user(self):
         with pytest.raises(Exception):
             user_is_group('doesnotexist')
+
+
+@pytest.mark.parametrize('user,group,is_in', [
+    ('ckuehl', 'ocfroot', True),
```
chriskuehl: idk, the test account tests seem sufficient. as much as possible it probably makes sense to avoid real usernames in tests (i know some exist already, but using test accounts seems preferable) as this will break some day.

ocflib/account/search.py
```diff
@@ -70,3 +71,16 @@ def user_is_group(username):
     """Returns whether username is an OCF group account."""
     attrs = user_attrs(username)
     return 'callinkOid' in attrs
+
+
+def users_in_group(groupname):
+    """Returns the list of users in an LDAP group."""
```
chriskuehl: Probably fine, but I feel this interface could confuse folks as it won't work for the user's primary gid.

For example, I could imagine some new staffer doing `users_in_group('sorry')` and wondering why it doesn't work.

I guess I don't have a strong opinion regarding whether this should be powered by LDAP or by local lookup.

tests/account/search_test.py
kpengboy: inb4 PEP 484

kpengboy: > ```python
> assert (user in users_in_group(group)) is is_in
> ```
Now practice saying that really fast...

kkuehler: Maybe change `is_in` to `member` or `in_group`?

dkess:
Closing this, since it's better to just rename the existing `list_staff` function.

PR #129, opened 2018-07-12
Rename list_staff and is_staff to be more general

dkess:


ocflib/account/utils.py
```diff
@@ -75,11 +77,29 @@ def is_staff(user, group='ocfstaff'):
...
+def is_in_group(user, group):
+    """Return whether the user is in a group.
+
+    :param group: UNIX group to use.
+    """
+    return user in list_group(group)
+
+
+def list_group(group):
+    """Return a list of OCF users in a group
```
kpengboy: I would clarify this to say that only returns users who have that group as a supplementary group. Or else, as Chris mentioned, someone might be confused at why `list_group('sorry')` gives you an empty list.

ocflib/account/utils.py
```diff
@@ -75,11 +77,29 @@ def is_staff(user, group='ocfstaff'):
 def list_staff(group='ocfstaff'):
```
kpengboy: Another option is to make this a wrapper around `is_in_group` (and deprecate just the second parameter). I don't have strong opinions either way.

ocflib/account/utils.py
```diff
@@ -87,6 +87,9 @@ def list_staff(group='ocfstaff'):
 def is_in_group(user, group):
     """Return whether the user is in a group.
 
+    Only will return True if the user is a supplementary member, so "sorry"
```
kpengboy: Well neither will `ocf` for that matter.

Why not say "Return whether the user is in the given supplementary group"? That way the first sentence is not so deceptive.

dkess: There's no such thing as a "supplementary group," a group is only supplementary if it's not a specific user's primary group.

kpengboy: I would still go with "Return whether the user is a supplementary member of the given group", except "supplementary member" isn't actually a term. And "Return whether the user has the given group among its supplementary group memberships" is too wordy. There doesn't seem to be a good way to phrase this...

Also I would've said "groups like 'sorry'" to avoid implying that `sorry` is the only group that won't work with this function.

PR #130, opened 2018-07-17
Add gapps_lists module

dkess:
This adds tools for interacting with our groups on GApps using the Google Admin API, and should lay the groundwork for eventually syncing LDAP groups and Google Groups.

One thing I'm not sure about in this commit is the way I handle authentication. My current plan is to have the credentials file get saved in the Puppet private share, and then have scripts provide that filename when constructing a `GAppsAdmin` object. But if there's a more idiomatic way to do this, I'd like to change it.

### DIFF
--- a/ocflib/misc/gapps_lists.py
+++ b/ocflib/misc/gapps_lists.py
@@ -1,0 +1,49 @@
+"""Google Admin Directory API for some mailing lists"""
+import googleapiclient.discovery
+from google.oauth2 import service_account
+
+
+class GAppsAdmin:
+    def __init__(self, service_account_file_path):
+        scopes = [
+            'https://www.googleapis.com/auth/admin.directory.group.member',
+        ]
+        credentials = service_account.Credentials.from_service_account_file(
+            service_account_file_path,
+            scopes=scopes,
+        )
+        delegated_credentials = credentials.with_subject(
+            'ocfbot@ocf.berkeley.edu'
+        )
+
+        self.groupadmin = googleapiclient.discovery.build(
+            'admin',
+            'directory_v1',
+            credentials=delegated_credentials,
+        )
+
+    def list_members(self, list_name):
+        """List all the OCF members (@ocf.berkeley.edu emails) in a GApps
+        mailing list. Strips email addresses, so this only returns email
+        addresses.
+
+        Ignores non-ocf.berkeley.edu emails and ocfbot@ocf.berkeley.edu.
+        """
+        response = self.groupadmin.members().list(groupKey=list_name).execute()
+        emails = (m['email'].split('@') for m in response['members'])
+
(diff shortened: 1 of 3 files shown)


abizer:
I wonder if it would be better to have this as a independent script in utils, for example, rather than adding another dependency to ocflib

dkess:
This won't ever get run by a human, so it didn't seem correct to put it in utils.

abizer:
That's true. However, it seems heavy to add two more dependencies that will get installed on all of our hosts for something that we will probably only want to run as a cronjob on a single host. Perhaps we should add an independent script to puppet instead?

ocflib/misc/gapps_lists.py
```diff
@@ -0,0 +1,49 @@
...
+        ]
+
+    def add_to_list(self, email, list_name):
+        """Adds an email address to a Google mailing list."""
+        body = {'email': email}
+
+        self.groupadmin.members().insert(
+            groupKey=list_name,
+            body=body,
+        ).execute()
```
kpengboy: We might want to also remove people from the list too.

kpengboy:
> One thing I'm not sure about in this commit is the way I handle authentication. My current plan is to have the credentials file get saved in the Puppet private share, and then have scripts provide that filename when constructing a GAppsAdmin object. But if there's a more idiomatic way to do this, I'd like to change it.

FYI the way we do this in things like ircbot is to have the constructor take a bunch of passwords as string arguments. (But if we move this to an independent script it would be a moot point.)

requirements-dev.txt
```diff
@@ -1,8 +1,10 @@
 coverage
 coveralls
 freezegun
+google-api-python-client
 ipdb
 mock
+oauth2client
```
chriskuehl: requirements-dev is not appropriate, these aren't dev dependencies; add to `setup.py` install_requires instead

chriskuehl:
+1 for keeping out of ocflib; this is logically more of a single-use application, so you could use dh-virtualenv, a service, etc.

unfortunately there is not a super straightforward way currently of having a "cronjob with python dependencies" besides dh-virtualenv (but there is prior art for this at the ocf)

dkess:
Sounds good. Which prior art are you referring to that uses dh-virtualenv?

chriskuehl:
https://github.com/ocf/approve is one example, not sure if we have others nowdays (create used to also work this way, but we turned it into a marathon service).

Would be nice if we had some kind of batch framework to use with mesos... not sure if we really want to invest in chronos at this point though, when the future of mesos at ocf seems unclear.

kpengboy:
On the subject of chronos, it's half-dead too.

dkess:
Closing, since this doesn't belong in ocflib.

PR #131, opened 2018-08-05
Switch to using a declarative Jenkinsfile

jvperrin:
This is much cleaner, hopefully it works the same way. I'm mostly concerned about some of the parallel stuff at the top, as I'm not sure if I've done it correctly. This is in a similar vein to ocf/dns#20, I eventually hope to have everything migrated to using this kind of format (and using the shared stuff in our [shared pipeline](https://github.com/ocf/shared-pipeline)).

I think this is probably the most complicated Jenkinsfile we have (due to all the parallel building stuff mostly), so after this is done I think migrating the other repos will be reasonably straightforward.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,94 +1,78 @@
-try {
-    node('slave') {
-        step([$class: 'WsCleanup'])
+def dists = ['jessie', 'stretch', 'buster']
 
-        stage('check-out-code') {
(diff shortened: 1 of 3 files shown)


jvperrin:
Seems like it has some problems with the parallel steps. Currently 1 fails while one succeeds, they're probably encountering some race condition between the two dists.

PR #134, opened 2018-08-19
Fix coverage when using pytest-xdist

jvperrin:
`pytest-cov` works with `pytest-xdist` whereas `coverage` does not, so this should fix the coveralls reports that we have 0% coverage and start reporting correct values again.


jvperrin:
Flaky DNS caused the build failures :(

PR #135, opened 2018-08-19
Use multiple DNS servers for lookups

jvperrin:
This should make tests more reliable and mean that tests fail less often due to one single DNS server not responding quickly.

I'm not a fan of needing to loop through DNS servers if a negative result is given, but I don't see a better way currently.

### DIFF
--- a/ocflib/infra/net.py
+++ b/ocflib/infra/net.py
@@ -11,6 +11,19 @@ OCF_SUBNET_V4 = ip_network('169.229.226.0/24')
 OCF_SUBNET_V6 = ip_network('2607:f140:8801::/48')
 OCF_SUBNET_V6_COMPAT = ip_network('2607:f140:8801::1:0/112')
 
+GOOGLE_PUBLIC_DNS_1 = ip_address('8.8.8.8')
+GOOGLE_PUBLIC_DNS_2 = ip_address('8.8.4.4')
+CLOUDFLARE_DNS_1 = ip_address('1.1.1.1')
+CLOUDFLARE_DNS_2 = ip_address('1.0.0.1')
+
+PUBLIC_DNS_SERVERS = [
+    GOOGLE_PUBLIC_DNS_1,
+    GOOGLE_PUBLIC_DNS_2,
+    CLOUDFLARE_DNS_1,
+    CLOUDFLARE_DNS_2,
+]
+DNS_SERVERS = [OCF_DNS_RESOLVER] + PUBLIC_DNS_SERVERS
+
 
 def ipv6_to_ipv4(ipv6):
     """Convert an OCF IPv6 address to its equivalent IPv4.
--- a/ocflib/misc/validators.py
+++ b/ocflib/misc/validators.py
@@ -4,7 +4,7 @@ import re
 import dns
 import dns.resolver
 
-from ocflib.infra.net import OCF_DNS_RESOLVER
+from ocflib.infra.net import DNS_SERVERS
 
 
 def host_exists(host):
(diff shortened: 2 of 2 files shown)


dkess:
If the issue is that DNS servers aren't responding quickly enough, how will this fix it? Looks like the DNS queries still have infinite timeouts.

jvperrin:
The issue is more that our DNS server (pestilence) returns `SERVFAIL` I believe (or maybe `NXDOMAIN`, not exactly sure). Either way, it says the domain doesn't exist when it really does, so that's the point of having multiple other DNS servers to try.

ocflib/misc/validators.py
```diff
@@ -13,7 +13,16 @@ def host_exists(host):
...
+    # through resolvers here instead and exit once one confirms the records
+    # actually exist (not great for lookups that return no records though):
+    # https://github.com/rthalley/dnspython/issues/117
+    for server in DNS_SERVERS:
+        response = dns.query.udp(message, str(server))
+
+        if bool(response.answer):
+            break
+
     return bool(response.answer)
```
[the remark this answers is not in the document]
jvperrin: Yeah, good point on that, it seemed to work to me, but I agree it looks out of scope by this point, so I'm not sure how that worked.

Anyway, the issue is that we can't return inside the loop when the response is coerced to `False`, because that could happen because the DNS server doesn't answer, or fails in some way. I think instead of doing what I did here, a better approach would be to check how if the response failed somehow (not sure what dnspython provides for that, but it probably has something) and then take action on that. That would also be better because then we don't have to contact like 5 different DNS servers to ever return `False` here...

I'll work on this a bit more before shipping it.

ocflib/misc/validators.py
```diff
@@ -13,7 +13,16 @@ def host_exists(host):
     except dns.name.EmptyLabel:
         return False
 
-    response = dns.query.udp(message, str(OCF_DNS_RESOLVER))
+    # Since dnspython resolvers don't support making ANY requests, we just loop
+    # through resolvers here instead and exit once one confirms the records
+    # actually exist (not great for lookups that return no records though):
+    # https://github.com/rthalley/dnspython/issues/117
+    for server in DNS_SERVERS:
+        response = dns.query.udp(message, str(server))
```
kkuehler: If our dns server is down, won't this just hang forever if we don't set a timeout?

ocflib/misc/validators.py
kkuehler: hmm... would the timeout param not handle this? in the loop you can return on true and keep iterating in the case of a timeout. return false if the loop is completed

ocflib/misc/validators.py
jvperrin: Hmmm, it looks like it would currently, the problem we are having right now I believe is not that it's down through, but that it instead returns some failure instead of giving the correct domain information.

Agreed that this should have a timeout specified though in case our DNS server (or whatever server we are contacting) happens to not respond.

ocflib/infra/net.py
```diff
@@ -11,6 +11,19 @@
 OCF_SUBNET_V6 = ip_network('2607:f140:8801::/48')
 OCF_SUBNET_V6_COMPAT = ip_network('2607:f140:8801::1:0/112')
 
+GOOGLE_PUBLIC_DNS_1 = ip_address('8.8.8.8')
```
abizer: mostly a nit, but why are we making these `ip_address`es when we are just going to cast them to str anyways? seems like it would be simpler to do 
```python
DNS_SERVERS = [
    str(OCF_DNS_RESOLVER),
    '8.8.8.8'  # google public DNS 1
    '8.8.4.4'  # google public DNS 2
    '1.1.1.1'  # cloudflare DNS 1
    '1.0.0.1'  # cloudflare DNS 2
]
```

jvperrin: I mostly did this because other IPs in that file are done in that way instead of as strings, and because then we could use the IPs in other places. Also, adding them as `ip_address`es adds more validation around them, but I don't think in this case it really matters.

jvperrin:
I'm resolving this as ocflib tests haven't been failing due to this recently, and we should be mocking this or something in tests anyways. That being said, adding some retries or additional servers in here would still probably be a good idea to avoid any flaky behavior.

PR #140, opened 2018-10-01
Monkeypatch out checking for affiliation, as berkeleyEduAffiliations …

redplanks:
…is being [removed](https://calnetweb.berkeley.edu/calnet-technologists/ldap-directory-service/changes-anonymous-ldap-binds-1012018) from anonymous LDAP searches.

please let me know if i missed anything for account creation that uses this attribute

### DIFF
--- a/ocflib/vhost/web.py
+++ b/ocflib/vhost/web.py
@@ -87,7 +87,8 @@ def eligible_for_vhost(user):
         return True
     elif 'calnetUid' in attrs:
         attrs_ucb = user_attrs_ucb(attrs['calnetUid'])
-        if attrs_ucb and 'EMPLOYEE-TYPE-ACADEMIC' in attrs_ucb['berkeleyEduAffiliations']:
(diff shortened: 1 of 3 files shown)


ocflib/vhost/web.py
```diff
@@ -87,7 +87,8 @@ def eligible_for_vhost(user):
         return True
     elif 'calnetUid' in attrs:
         attrs_ucb = user_attrs_ucb(attrs['calnetUid'])
-        if attrs_ucb and 'EMPLOYEE-TYPE-ACADEMIC' in attrs_ucb['berkeleyEduAffiliations']:
+        # TODO: Uncomment when we get a privileged LDAP bind.
+        if attrs_ucb:  # and 'EMPLOYEE-TYPE-ACADEMIC' in attrs_ucb['berkeleyEduAffiliations']:
```
kpengboy: This will permit arbitrary members of the OCF to submit vhost requests. You ought to be prepared for that.

Issue #141, opened 2018-10-02
migrate ocflib.infra.db to ocflib.infra.mysql


abizer:
as we now have a postgres service, we should rename ocflib.infra.db to ocflib.infra.mysql. it would be a good time to take care of things like #108  as well

MCer4294967296:
A naive question that might be of poor formatting: after both using `find . -type f -exec grep -H 'db' {} \;` in my cloned `ocflib` repo and searching on github with criteria `org:ocf ocflib.infra db`, there are only three other files referencing that "db.py" file. Does it solve this issue if I change the file name to "mysql.py" and change the references in the three files from "db" to "mysql"?

abizer:
That should work, but you'll also need to migrate any direct calls to db.get_connection in any downstream projects like ocfweb, puppet, utils, etc, as well as in the tests, if there are any. I think there are a few.

PR #147, opened 2018-11-01
migrate ocflib.infra.db to ocflib.infra.mysql

MCer4294967296:
solving issue #141. Thorough search tells me there really is not a lot of references to `db.py`... There are some more references in the Decal repo though, which is why I have not deleted `db.py`.

btw I am Billy, ocf username `yichizh`


dkess:
I would make it more clear in the comment that `ocflib.infra.db` is deprecated and shouldn't be used.

ethanhs:
>include a print statement that says it's been deprecated

I believe a `DeprecationWarning` might be better suited for this?

abizer:
DeprecationWarning is even better

ocflib/infra/db.py
```diff
@@ -1,5 +1,8 @@
 import pymysql
+import warnings
 
+warn("ocflib.infra.db is deprecated, use ocflib.infra.mysql instead.",
```
kpengboy: The correct way to do this is `warnings.warn`. You may also want to specify `stacklevel=2`.

ethanhs:
Wait why are we deleting this file that we had put a deprecation warning on?

abizer:
the deprecation warning is actually superfluous since the rename is contained entirely inside of ocflib. all external uses go through a database-specific interface

abizer:
@MCer4294967296 bump

PR #148, opened 2018-11-03
Allow LDAP changes without a keytab

jvperrin:
This then allows these functions to be used in user-facing scripts where they have potentially already authenticated with their password and have an existing Kerberos ticket. (like in `update-email` and `chsh`, the [two remaining python2 scripts](https://github.com/ocf/utils/issues/19) we have)

I also did a pretty major refactor to use `subprocess` instead of `pexpect` for changing LDAP since ldapmodify can just accept input from stdin instead of needing it to be passed in interactively. I also removed a celery task for modifying LDAP attributes that isn't used anywhere outside of ocflib as far as I could tell with sourcegraph.

I tested this manually with `./tests-manual/infra/create-ldap-keytab` and that worked fine, along with manually testing in a console that I could change my `loginShell` and `mail` attributes without a keytab being passed in.

### DIFF
--- a/ocflib/infra/ldap.py
+++ b/ocflib/infra/ldap.py
@@ -1,4 +1,5 @@
 import string
+import subprocess
 from base64 import b64encode
 from contextlib import contextmanager
 from datetime import datetime
@@ -6,7 +7,6 @@ from itertools import chain
 from textwrap import dedent
 
 import ldap3
-import pexpect
 
 from ocflib.misc.mail import send_problem_report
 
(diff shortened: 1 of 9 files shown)


gnowxilef:
i'm curious, is there no python library thingy that can manipulate ldap than writing a bunch of custom modify ldap stuffs?

jvperrin:
There is, but I haven't found anything that works with Python 3, LDAP, and Heimdal Kerberos all in one package. Plus we already do this for account creation, so I figured using it elsewhere would make sense.

gnowxilef:
ah, bummer :(

ocflib/infra/ldap.py
```diff
@@ -89,90 +89,107 @@ def format_value(value):
...
     :param attributes: dict mapping attribute name to list of values
-    :param keytab: path to the admin keytab
-    :param admin_principal: admin principal to use with the keytab
+    :param **kwargs: any additional keyword arguments to pass on to _write_ldif
     """
     lines = chain(
         _format_attr('dn', [dn]),
         ('changetype: add',),
-        *(_format_attr(key, values) for key, values in attributes.items())
+        *(_format_attr(key, values) for key, values in sorted(attributes.items()))
```
kpengboy: What was the impetus for `sorted`?

jvperrin: Ah, that's a good question. This was added so that the tests could check for an exact output here instead of doing something messier with it to protect against potential reordering. I also figured the number of attributes would be pretty low, so sorting wouldn't have any measurable impact on the performance of this or anything.

PR #154, opened 2019-01-30
ocflib shorturls db interface

abizer:


dkess:
Perhaps we should move functions that access the shorturl database from ircbot to ocflib?

ocflib/misc/ocfshorturls.sql
```diff
@@ -0,0 +1,82 @@
...
+-- --------------------------------------------------------
+
+--
+-- Table structure for table `shorturls`
+--
+
+CREATE TABLE `shorturls` (
+  `id` int(11) NOT NULL,
+  `time_created` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `slug` varchar(100) NOT NULL,
```
kpengboy: I thought the maximum was 255.

abizer: mariadb doesn't allow you to specify a unique constraint on columns that wide so I cut it down. a 255 character long shorturl doesn't make sense as it is, this is some arbitrary smaller value

abizer: which collation would you suggest? I'd like to keep this simple lol

ocflib/misc/shorturls.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def delete_shorturl(ctx, slug):
+    """Delete a shorturl from the database."""
+
+    query = 'DELETE FROM `shorturls` WHERE `slug` = %s'
+    ctx.execute(query, (slug,))
+
+
+def rename_shorturl(ctx, old_slug, new_slug):
```
dkess: Don't remember if I ever expressed this elsewhere, but I don't believe we should have this rename function. It seems like it doesn't have a strong use case anyways, and if it's really needed we can just delete and then re-add.

PR #155, opened 2019-02-01
Add new hours2 api

dkess:
This adds a new API for getting lab hours. The plan is to deprecate the old API and replace it with this one. In the meantime there will have to be an "`hours`" and "`hour2`".

This API is a bit nicer because it stores the entire hours listing in a single object, and methods on that object work on the entire calendar. This is better than our current system since it makes it easier to cache hours for ocfweb.

This also reads hours from a yaml file (in etc) instead of being part of the library itself. I also used the `attrs` library for nice validation and conversion from YAML to object types.

### DIFF
--- a/ocflib/lab/hours2.py
+++ b/ocflib/lab/hours2.py
@@ -1,0 +1,289 @@
+"""Methods for dealing with OCF lab hours.
+
+All times are assumed to be OST (OCF Standard Time).
+
+Usage:
+
+    >>> from ocflib.lab.hours2 import read_hours_listing()
+    >>> hours_listing = read_hours_listing()
+    >>> hours_listing.hours_on_date(date(2015-10-12))
+    [Hour(open=datetime.time(9, 0), close=datetime.time(18, 0))]
+"""
+from datetime import date
+from datetime import datetime
+from datetime import time
+from datetime import timedelta
+from enum import IntEnum
+
+import attr
+import yaml
+
+
+class Weekday(IntEnum):
+    Monday = 0
+    Tuesday = 1
+    Wednesday = 2
+    Thursday = 3
+    Friday = 4
+    Saturday = 5
+    Sunday = 6
+
+
+def _parsetime(t):
+    if isinstance(t, time):
+        return t
+
+    return datetime.strptime(t, '%H:%M').time()
+
+
+def _parse_regular_hours(regular):
+    """Parses a dictionary of regular hours, as used in the yaml file.
+
+    Keys can be either weekday strings ("Monday", "Tuesday", ...) or Weekday
+    enums. Values should be a list of timeranges, parseable by
+    _parse_hours_list.
+    """
+    out = {}
+    for weekday, hours in regular.items():
+        try:
+            # First see if it's an int already
+            weekday = Weekday(weekday)
+        except ValueError:
+            # Convert strings ('Monday') to int (0)
+            weekday = getattr(Weekday, weekday)
+
+        out[weekday] = _parse_hours_list(hours)
+
+    if out.keys() != set(Weekday):
+        raise ValueError('Regular hours must be set for all weekdays')
+
+    return out
+
+
+def _parse_hours_list(time_ranges):
+    """Converts a list of timeranges to a list of Hour objects.
+
+    This function idempotently does nothing on a list of Hour objects.
+
+    Raises an ValueError if the timeranges overlap.
+
+    >>> _parse_hours_list([['8:30', '10:00'], ['11:00', '3:00']])
+    [Hour(time(8, 30), time(10)), Hour(time(11), time(3))]
+
+    >>> _parse_hours_list([['8:30', '10:00'], ['9:00', '11:00']])
+    Traceback (most recent call last):
+        ...
+    ValueError: Hours must be in order
+
+    >>> _parse_hours_list([['8:30', '10:00'], ['8:00', '11:00']])
+    Traceback (most recent call last):
+        ...
+    ValueError: Hours must be in order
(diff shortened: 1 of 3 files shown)


ocflib/lab/hours2.py
```diff
@@ -0,0 +1,289 @@
+"""Methods for dealing with OCF lab hours.
+
+All times are assumed to be OST (OCF Standard Time).
+
+Usage:
+
+    >>> from ocflib.lab.hours import read_hours_listing()
```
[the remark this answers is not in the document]
dkess: I guess we could do this, but since I expect the decommission process to be very quick (this isn't used in many parts of the code) I don't think it matters to have this kind of separation.

To elaborate, I expect the deprecation process to look like this:

1. commit this code
2. change all code to use the new library
3. decommission the old library and make it an alias for the new library
4. remove the `hours2` name

ocflib/lab/hours2.py
```diff
@@ -0,0 +1,289 @@
...
+
+    return out
+
+
+def _parse_hours_list(time_ranges):
+    """Converts a list of timeranges to a list of Hour objects.
+
+    This function idempotently does nothing on a list of Hour objects.
+
+    Raises an ValueError if the timeranges overlap.
```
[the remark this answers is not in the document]
dkess: If they overlap then they also won't be in order, but perhaps I can change the comment to make that more clear.

ocflib/lab/hours2.py
```diff
@@ -0,0 +1,289 @@
...
+    ValueError: Hours must be in order
+    """
+    hours = []
+
+    for hour in time_ranges:
+        if not isinstance(hour, Hour):
+            hour = Hour(hour[0], hour[1])
+
+        if hours and hours[-1].close > hour.open:
+            raise ValueError('Hours must be in order')
```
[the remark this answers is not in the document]
dkess: This makes the logic for checking overlaps a bit easier, but more importantly I think we should error on this because the configuration files shouldn't list hours or holidays out of order, and our validation should alert us when this happens. It's probably a symptom of a broken copy-paste and possibly something else being broken.

ocflib/lab/hours2.py
```diff
@@ -0,0 +1,289 @@
...
+
+    Keys can be either weekday strings ("Monday", "Tuesday", ...) or Weekday
+    enums. Values should be a list of timeranges, parseable by
+    _parse_hours_list.
+    """
+    out = {}
+    for weekday, hours in regular.items():
+        try:
+            # First see if it's an int already
+            weekday = Weekday(weekday)
```
ethanhs: Please use `isinstance(weekday, int)` to check this.

ocflib/lab/hours2.py
```diff
@@ -0,0 +1,289 @@
...
+class HoursListing(object):
+    regular = attr.ib(converter=_parse_regular_hours)
+    holidays = attr.ib(converter=_parse_holiday_list)
+
+    def hours_on_date(self, when=None):
+        """Returns the hours on the given date.
+
+        If not provided, when defaults to today.
+        """
+        if not when:
```
ethanhs:
```suggestion
        if when is None:
```

ocflib/lab/hours2.py
```diff
@@ -0,0 +1,289 @@
...
+                return holiday.hours
+
+        return self.regular[when.weekday()]
+
+    def is_open(self, when=None):
+        """Return whether the lab is open at the given datetime.
+
+        If not provided, when defaults to now.
+        """
+        if not when:
```
ethanhs:
```suggestion
        if when is None:
```

ocflib/lab/hours2.py
```diff
@@ -0,0 +1,289 @@
...
+        )
+
+    def time_to_open(self, when=None):
+        """Return timedelta object representing time until the lab is open from the given datetime.
+
+        If the lab will never be open, returns None.
+
+        If not provided, when defaults to now.
+        """
+        if not when:
```
ethanhs:
```suggestion
        if when is None:
```

ocflib/lab/hours2.py
```diff
@@ -0,0 +1,289 @@
...
+        return None
+
+    def time_to_close(self, when=None):
+        """Return timedelta object representing time until the lab is closed from the given datetime.
+
+        If the lab will never close, returns None.
+
+        If not provided, when defaults to now.
+        """
+        if not when:
```
ethanhs:
```suggestion
        if when is None:
```

ocflib/lab/hours2.py
```diff
@@ -0,0 +1,289 @@
...
+    if isinstance(holiday, Holiday):
+        return holiday
+
+    if isinstance(holiday['date'], list):
+        start, end = holiday['date']
+    elif isinstance(holiday['date'], date):
+        start = end = holiday['date']
+    else:
+        raise ValueError(
+            'Holiday date is not a (start, end) pair or single datetime'
```
ethanhs: nit: if we expect a list and not a tuple, you should say `is not a [start, end] pair`
```suggestion
            'Holiday date is not a [start, end] pair or single datetime'
```

ocflib/lab/hours2.py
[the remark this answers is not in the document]
jvperrin: I think that would be good, right now it's not clear from this comment that this also requiring the hours passed in to be in order (except the examples shown).

The error message also makes it clear that they should be in order, but not that overlaps are forbidden (in my opinion, being in order just means that all the start times are in order, not that there are no overlaps, so being clear about that would be good).

PR #158, opened 2019-02-16
Decommission old hours library

dkess:
This PR decommissions the old library and puts the new one in its place. This will break linters for a bit because of how it's done, but once we migrate everything over to the new name this will look normal again.

(though if someone knows a better way of doing this without making the linters angry please lmk)


jameslzhu:
Would it be better to print a deprecation warning, instead of wholesale deletion?

https://docs.python.org/3/library/warnings.html#warnings.warn

Alternatively, keeping the API but throwing an exception on import may work out better.

I don’t know how much breakage we want (this may be desirable) or whether throwing an error at import or runtime is desirable.

dkess:
I did a [sourcegraph search](https://sourcegraph.ocf.berkeley.edu/search?q=ocflib.lab.hours) to make sure that the old library isn't used anywhere else. While I suppose it's possible that it's lingering in some other places, I'm not sure if a `DeprecationWarning` would be visible to us in those places. Failure is more visible, so I'd rather them straight up error out so we can find out where they are.

PR #159, opened 2019-03-04
Add last used function

TonyLianLong:
I added a last used function since it would be useful if we can figure out who is the last user in cases such as (https://github.com/ocf/utils/issues/96).

We can use `last_used(...)["end"] == None` to determine whether the user is currently using the machine.


dkess:
`ocflib.lab.stats` uses the anonymous MySQL user, whereas this uses credentials, so I think it's OK for this to be in a separate module.

abizer:
the functions in that class have the ability to take credentials. splitting this functionality out like this without good reason breaks the logical design of ocflib.

TonyLianLong:
So probably I can put the code into that module and refractor a little bit so that when specified with some arguments it does not connect with anonymous username?

dkess:
Look at how the other functions call `get_connection`. It's actually a partial function, which means you can call `get_connection(user='username', password='aaaaa', db='db')`, etc.

TonyLianLong:
Now I put those functions into stats.py with some slight modifications.

ocflib/lab/stats.py
```diff
@@ -139,6 +139,36 @@ def list_desktops(public_only=False):
         c.search(OCF_LDAP_HOSTS, filter, attributes=['cn'])
         return [entry['attributes']['cn'][0] for entry in c.response]
 
+def build_query(host):
```
dkess: Since this function is only used in `last_used`, I'd take it out of its own separate function and combine it with `last_used`.

ocflib/lab/stats.py
```diff
@@ -139,6 +139,36 @@ def list_desktops(public_only=False):
...
+    def add_ocf(hostname):
```
dkess: This can be in its own function. It shouldn't be nested, though. It's fine for this to be in global scope.

ocflib/lab/stats.py
```diff
@@ -139,6 +139,36 @@ def list_desktops(public_only=False):
...
+    query += ' ORDER BY `start` DESC LIMIT 1'
+    # we can't have another user start before current user ends so here we order by start
+
+    return (query, query_args)
+
+def last_used(desktop_host):
+    try:
+        with open('/etc/ocfstats-ro.passwd', 'r') as fin:
+            password = fin.read().strip()
+    except FileNotFoundError:
```
dkess: No need to catch this error, I'd just let the function fail.

dkess:
This lgtm, to fix the last errors you should make sure to run `make install-hooks` on your local environment. Then running `git commit` will automatically fix the style errors. You can also run `pre-commit` manually to fix the errors.

ocflib/lab/stats.py
```diff
@@ -140,6 +140,34 @@ def list_desktops(public_only=False):
...
+
+        query = 'SELECT * FROM `session`'
+        query_args = []
+
+        query_conditions = []
+        query_conditions.append('`host` = %s')
+        query_args.append(add_ocf(host))
+
+        query += ' WHERE ' + ' AND '.join(query_conditions)
+        query += ' ORDER BY `start` DESC LIMIT 1'
```
abizer: while this query building is flexible, it's rather overengineered for the usecase it's in: we're only adding a single condition anyways. To make it easier to grok for future people reading this code, it would be better to rewrite it explicitly:

```python

query = 'SELECT * FROM `session` WHERE `host` = %s ORDER BY `start` DESC LIMIT 1'
c.execute(query, host)
return Session.from_row(c.fetchone())
```

abizer:
There are still issues with this PR that ought to be fixed.

ocflib/lab/stats.py
```diff
@@ -140,6 +140,34 @@ def list_desktops(public_only=False):
...
+        return hostname + '.ocf.berkeley.edu'
+    return hostname
+
+
+def last_used(host):
+    with open('/etc/ocfstats-ro.passwd', 'r') as fin:
+        password = fin.read().strip()
+    with mysql.get_connection(user='ocfstats-ro', password=password, db='ocfstats') as c:
+        # The reason for not using the partial `get_connection` is because of this:
+        # https://github.com/PyCQA/pylint/issues/2271
```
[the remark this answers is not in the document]
TonyLianLong: `# noqa` cannot disable error-checking on this error, and thus I used `pylint: disable=E1124`

PR #163, opened 2019-03-10
Improve on last used function

TonyLianLong:
This continues the PR #159 because Abizer has made useful suggestions to `last_used`.

### DIFF
--- a/ocflib/lab/stats.py
+++ b/ocflib/lab/stats.py
@@ -8,6 +8,7 @@ from datetime import timedelta
 from cached_property import cached_property
 
 from ocflib.infra import mysql
+from ocflib.infra.hosts import domain_from_hostname
 from ocflib.infra.ldap import ldap_ocf
 from ocflib.infra.ldap import OCF_LDAP_HOSTS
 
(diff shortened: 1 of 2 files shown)


ocflib/lab/stats.py
```diff
@@ -140,6 +141,17 @@ def list_desktops(public_only=False):
         return [entry['attributes']['cn'][0] for entry in c.response]
 
 
+def last_used(hostname, password):
+    """Show the last used statistics for a computer."""
+    with get_connection(user='ocfstats-ro', password=password) as c:  # pylint: disable=E1124
```
abizer: the signature for this function should be

```python
def last_used(hostname, ctx):
  query = "..."
  ctx.execute(...)
  return Session.from_row(c.fetchone())
```

i.e. build the MySQL connection object in the caller, not in the callee.

ocflib/lab/stats.py
```diff
@@ -140,6 +141,17 @@ def list_desktops(public_only=False):
...
+
```
abizer: nit: unnecessary whitespace

ocflib/lab/stats.py
```diff
@@ -172,12 +184,8 @@ def from_hostname(cls, hostname, start, end):
 
     @classmethod
     def from_hostnames(cls, hostnames, start, end):
-        def add_ocf(hostname):
-            if not hostname.endswith('.ocf.berkeley.edu'):
-                return hostname + '.ocf.berkeley.edu'
-            return hostname
 
-        hostnames = tuple(map(add_ocf, hostnames))
+        hostnames = tuple(map(domain_from_hostname, hostnames))
```
abizer: why do we call `domain_from_hostname` redundantly here and above? IMO any data we are passing into deeper layers of code should get sanitized at higher layers

ocflib/lab/stats.py
TonyLianLong: Sure, and in this case we don't have to consider the bug in pylint.

TonyLianLong: But then it would be different from other functions in this file, which calls `get_connection`, and does it matter? Since other functions do not need external context to authenticate, I guess it's fine?

abizer: it's fine, though in some cases we may want to consider updating their signatures as well. Depends on where they're used.

Issue #162, opened 2019-03-10
Add kanboard support


dkess:
See [ocflib.infra.discourse](https://github.com/ocf/ocflib/blob/master/ocflib/infra/discourse.py) for inspiration on how to do this. I believe users can make personal Kanboard tokens for this (for testing purposes), and ircbot can use the global admin token.

This is necessary for https://github.com/ocf/ircbot/issues/106

willtunnels:
Looks like the Kanboard API uses JSON-RPC 2. We have a couple options:
(1) add the official kanboard python client as a dependency (might be overkill unless we plan on adding more kanboard functionality)
(2) do an ad hoc request with the right format
(3) add a generic JSON-RPC library as a dependency (probably not)

dkess:
Their documentation has examples that use `curl`, so I assume it won't be too hard for us to make ad-hoc requests with the right format (using the `requests` library).

cg505:
I believe this should have been closed.

PR #164, opened 2019-03-15
Add support for querying Kanboard tasks. Fixes issue #162

willtunnels:


dkess:
What is `usr` supposed to mean here? I tried running the manual test with my personal API key and username and I got a 401 error

willtunnels:
`usr` is supposed to be the username you use to login to Kanboard. It worked for me and I don't have admin privileges or anything. If you do have admin privileges it's possible you actually used the application key (which you can see as an admin) instead of your user key. Your user key should be under My profile -> Actions -> API. If you've never explicitly clicked a button to generate your key, it's the application key. The application key has to be used with usr="jsonrpc" and that's what the irc bot should do. See [the docs](https://docs.kanboard.org/en/latest/api/introduction.html). If that's not the problem idk what's happening.

abizer:
add some docstrings if possible but otherwise this lgtm

ocflib/infra/kanboard.py
```diff
@@ -34,6 +38,19 @@ def __str__(self):
 
     @classmethod
     def from_number(cls, usr, api_key, num):
+        """Gets information about a Kanboard task based on its number.
+
+        Example usage:
+
+            KanboardTask.from_number('jsonrpc',
+                '19ffd9709d03ce50675c3a43d1c49c1ac207f4bc45f06c5b2701fbdf8929', 1)
```
dkess: Is this a real API key?

willtunnels: No, it's not. I copied it from an example on the Kanboard API docs.

PR #165, opened 2019-03-18
account: reserve some UIDs in the 60000 range

### LINKED ISSUE #127: Reserve UIDs in 60000 range
There are some UIDs that we better not hand out to new users:

 - 65534 is the `nobody` user
 - 65535 is sometimes used as an invalid user
 - 61184-65519 is reserved by systemd for [dynamic users](http://0pointer.net/blog/dynamic-users-with-systemd.html)

dkess:
Fixes #127. I'm pretty sure this works but it's possible I missed something, so a close reading from someone else would be helpful.

### DIFF
--- a/ocflib/account/creation.py
+++ b/ocflib/account/creation.py
@@ -49,6 +49,11 @@ JDNjiYba1ZiNLiqXeLGS2IVYAd88etX+V5gxAvl0bGHzgeHodutxUf46QCg7cmvm
 zwIDAQAB
 -----END PUBLIC KEY-----'''
 
+RESERVED_UID_RANGES = [
+    (61184, 65519),  # systemd dynamic users
+    (65534, 65535),  # nobody and invalid user
+]
+
 
 def _get_first_available_uid(known_uid=_KNOWN_UID):
     """Return the first available UID number.
@@ -72,7 +77,20 @@ def _get_first_available_uid(known_uid=_KNOWN_UID):
     else:
         # If cached UID is later deleted, LDAP response will be empty.
         max_uid = known_uid
-    return max_uid + 1
+
+    assert all(start <= end for start, end in RESERVED_UID_RANGES)
+    next_uid = max_uid + 1
+    while True:
+        for start, end in sorted(RESERVED_UID_RANGES):
+            if start <= next_uid <= end:
+                next_uid = end + 1
+                break
+        else:
+            # The else clause is run if we never break out of the loop
+            # (next_uid is not in any reserved ranges)
+            break
+
+    return next_uid
 
 
 def create_account(request, creds, report_status, known_uid=_KNOWN_UID):
--- a/ocflib/account/creation.py
+++ b/ocflib/account/creation.py
@@ -49,9 +49,13 @@ JDNjiYba1ZiNLiqXeLGS2IVYAd88etX+V5gxAvl0bGHzgeHodutxUf46QCg7cmvm
 zwIDAQAB
 -----END PUBLIC KEY-----'''
 
+# Inclusive endpoints
 RESERVED_UID_RANGES = [
-    (61184, 65519),  # systemd dynamic users
-    (65534, 65535),  # nobody and invalid user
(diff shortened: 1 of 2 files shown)


ocflib/account/creation.py
```diff
@@ -49,6 +49,11 @@
 zwIDAQAB
 -----END PUBLIC KEY-----'''
 
+RESERVED_UID_RANGES = [
+    (61184, 65519),  # systemd dynamic users
+    (65534, 65535),  # nobody and invalid user
```
abizer: IMO we should just close off the entire range, it's only 15 extra uids anyways

ocflib/account/creation.py
```diff
@@ -72,7 +77,20 @@ def _get_first_available_uid(known_uid=_KNOWN_UID):
...
-    return max_uid + 1
+
+    assert all(start <= end for start, end in RESERVED_UID_RANGES)
+    next_uid = max_uid + 1
+    while True:
+        for start, end in sorted(RESERVED_UID_RANGES):
+            if start <= next_uid <= end:
+                next_uid = end + 1
+                break
+        else:
```
redplanks: Indentation here

ocflib/account/creation.py
```diff
@@ -72,7 +77,20 @@ def _get_first_available_uid(known_uid=_KNOWN_UID):
     else:
         # If cached UID is later deleted, LDAP response will be empty.
         max_uid = known_uid
-    return max_uid + 1
+
+    assert all(start <= end for start, end in RESERVED_UID_RANGES)
+    next_uid = max_uid + 1
+    while True:
```
kpengboy: Some comments:
 - Can we assume that `RESERVED_UID_RANGES` is in sorted order and disjoint? It would simplify the code quite a bit.
 - Even if we can't assume disjointedness, can't we still loop through that list only once? Why do we need the outer `while` loop?

ocflib/account/creation.py
```diff
@@ -49,6 +49,11 @@
 zwIDAQAB
 -----END PUBLIC KEY-----'''
 
+RESERVED_UID_RANGES = [
```
kpengboy: Please document whether the endpoints are inclusive or exclusive

ocflib/account/creation.py
redplanks: Nit: get rid of the `else`? There's not corresponding `if` clause

ocflib/account/creation.py
dkess: You're probably right that the `while` loop isn't necessary, I'll change that

ocflib/account/creation.py
dkess: This is the python for-else feature, though I think kpengboy is right that it might not be needed

dkess:
Thanks everyone for the feedback. I've simplified the code considerably.

We sort the ranges, so it doesn't matter if they're out of order. I believe the code will work even if the ranges overlap.

ocflib/account/creation.py
redplanks: Huh, TIL

tests/account/creation_test.py
```diff
@@ -101,6 +101,25 @@ def ldap_ocf():
 
         assert next_uid == 999201
 
+    def test_reserved_uid(self):
+        """Test that we skip over UIDs 61184-65519, the range of systemd
```
redplanks: Nit: update comment to 65535, or remove references to specific ranges entirely?

tests/account/creation_test.py
```diff
@@ -101,6 +101,25 @@ def ldap_ocf():
...
+        dynamic users.
+        """
+
+        connection = mock.Mock(response=[
+            {'attributes': {'uidNumber': 61183}},
+            {'attributes': {'uidNumber': 60000}},
```
redplanks: Add a test for this too? To check that it doesn't skip to the end of reserved ranges. Though that would probably overlap with other tests.

dkess: eh i think this is pretty much handled by the other tests

PR #157, opened 2019-02-12
Privileged ldap

ja5087:
addresses #142

PR #169, opened 2019-04-20
Privileged LDAP (again)

ja5087:
Last time we tried to merge this (see #157) the prerequisite ucbldap.passwd was not present in ocfweb, dev-ocfweb etc. 

I've rolled back all the changes I made to the tests. Once the password file has been added to the puppet configs for ocfweb and Jenkins, this PR should be good for merging.

### DIFF
--- a/ocflib/account/creation.py
+++ b/ocflib/account/creation.py
@@ -274,7 +274,7 @@ def validate_calnet_uid(uid):
         raise ValidationError(
             'CalNet UID already has account: ' + str(existing_accounts))
 
-    attrs = search.user_attrs_ucb(uid)
+    attrs = search.user_attrs_ucb_privileged(uid)
 
     if not attrs:
         raise ValidationError("CalNet UID can't be found in university LDAP.")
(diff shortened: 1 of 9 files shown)


cg505:
I think you want to take another look at this. `ocflib/account/search.py` was broken by the master merge. The master merge also included the revert of the last version, so you may need to include a commit reverting that as well. (For instance, `ocflib/infra/ldap.py` in this PR does not contain the `ldap_ucb_privileged` function. `git revert bec763573a74fad348e6c43254009bb79a1b472c` may do what you want, but you should manually verify that.)

ja5087:
well this is not ideal

cg505:
@ja5087 I added a commit. I think now you will just need to fix the tests to account for the different functions.

Sorry for all the trouble. Reverting merges in git is known to be annoying and hard to handle.

kpengboy:
What's the current status of this?

ja5087:
It's been stalled for a while, but I do have a better understanding of the system now and could get back on it soon

ethanhs:
@ja5087 would you have time to re-base and bring this over the finish line at some point? Would love to see this landed!

axmmisaka:
I can potentially finish this at some point if I still have time, but I'm unsure if password is used frequently as what I encountered mostly is that we use Kerberos ticket to auth???

PR #170, opened 2019-04-24
Reserve name "rachy"

dkess:
This is a custom mastodon pseudonym


cg505:
no longer needed, rachy is an account

PR #171, opened 2019-04-26
escape ldap3 filter chars

### LINKED ISSUE #150: LDAP search (ocflib.account.search) doesn't escape searches properly
This causes errors like this, when a search contains parentheses:

```
Traceback (most recent call last):
  File "/opt/ocfweb/venv/lib/python3.5/site-packages/django/core/handlers/base.py", line 124, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/ocfweb/ocfweb/account/register.py", line 122, in recommend
    recommendations = recommender.recommend(real_name, 10)
  File "/opt/ocfweb/ocfweb/account/recommender.py", line 30, in recommend
    validate_username(rec, real_name)
  File "/opt/ocfweb/venv/lib/python3.5/site-packages/ocflib/account/creation.py", line 328, in validate_username
    if search.user_exists(username):
  File "/opt/ocfweb/venv/lib/python3.5/site-packages/ocflib/account/search.py", line 61, in user_exists
    return bool(user_attrs(account))
  File "/opt/ocfweb/venv/lib/python3.5/site-packages/ocflib/account/search.py", line 48, in user_attrs
    c.search(base, '(uid={})'.format(uid), attributes=ldap3.ALL_ATTRIBUTES)
  File "/opt/ocfweb/venv/lib/python3.5/site-packages/ldap3/core/connection.py", line 779, in search
    check_names=self.check_names)
  File "/opt/ocfweb/venv/lib/python3.5/site-packages/ldap3/operation/search.py", line 372, in search_operation
    request['filter'] = compile_filter(parse_filter(search_filter, schema, auto_escape, auto_encode, validator, check_names).elements[0])  # parse the searchFilter string and compile it starting from the root node
  File "/opt/ocfweb/venv/lib/python3.5/site-packages/ldap3/operation/search.py", line 220, in parse_filter
    raise LDAPInvalidFilterError('invalid filter')
ldap3.core.exceptions.LDAPInvalidFilterError: invalid filter



Request:
  * Host: www.ocf.berkeley.edu
  * Path: /account/register/recommend/?real_name=Xiao%20(Calvin)%20Hu
  * Method: GET
  * Secure: True
```

We might be able to use `escape_bytes` from `ldap3` or maybe just escape parentheses and move on since we quite rarely encounter this issue.

dkess:
Resolves #150.

### DIFF
--- a/ocflib/account/search.py
+++ b/ocflib/account/search.py
@@ -1,5 +1,6 @@
 """Methods for searching and selecting users."""
 import ldap3
+from ldap3.utils.conv import escape_filter_chars
 
 import ocflib.infra.ldap as ldap
 from ocflib.infra.ldap import OCF_LDAP_PEOPLE
@@ -23,13 +24,17 @@ def users_by_filter(ldap_filter):
 def users_by_calnet_uid(calnet_uid):
     """Get a list of users associated with a CalNet UID"""
     calnet_uid = int(calnet_uid)
-    return users_by_filter('(calnetUid={})'.format(calnet_uid))
+    return users_by_filter(
+        '(calnetUid={})'.format(escape_filter_chars(calnet_uid))
+    )
 
 
 def users_by_callink_oid(callink_oid):
     """Get a list of users associated with a CalLink OID"""
     callink_oid = int(callink_oid)
-    return users_by_filter('(callinkOid={})'.format(callink_oid))
+    return users_by_filter(
+        '(callinkOid={})'.format(escape_filter_chars(callink_oid))
+    )
 
 
 def user_attrs(uid, connection=ldap.ldap_ocf, base=OCF_LDAP_PEOPLE):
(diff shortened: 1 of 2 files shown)


dkess:
I guess I could do something like that, but I don't really trust myself to write that method in a way that's safe and flexible. `ldap3` should really have this built-in...

kpengboy:
We could take inspiration from SQL prepared statements. I would still want to have some central function for executing an LDAP query with escaped arguments. True, we aren't experts and might miss some corner case. But the status quo solution of manually inspecting and patching every call to `connection.search` isn't any better in that respect, and runs the added risk of missing a patch somewhere.

dkess:
Bumping this because I've been seeing the issue repeat itself in rootspam a lot lately. Since we don't use this kind of code very much, I'm not convinced it's worth the effort to make the underlying function safe to these attacks. Thoughts?

ethanhs:
I think we should stop the bleeding, merge this and open an issue for follow up

dkess:
smash that approve button

PR #173, opened 2019-05-04
add `mastodon` to validators.py

MrMinos:

PR #174, opened 2019-05-04
Revert "add `mastodon` to validators.py"

MrMinos:
Reverts ocf/ocflib#173, the commit was failing to build


abizer:
this still breaks. probably try to figure out what the bug is and fix that, since reverting it doesn't appear to help.

PR #177, opened 2019-05-07
Happy Hour Spring '19

dongkyunc:


ocflib/printing/quota.py
```diff
@@ -19,8 +19,8 @@
 #     them permanent. If it doesn't, then those changes automatically expire.
 
 HAPPY_HOUR_QUOTA = 20
-HAPPY_HOUR_START = datetime(2018, 12, 3)
-HAPPY_HOUR_END = datetime(2018, 12, 16)
+HAPPY_HOUR_START = datetime(2019, 5, 6)
+HAPPY_HOUR_END = datetime(2019, 7, 17)
```
redplanks: Should be 5-17 or 5-16?

dongkyunc: why am i such a dumbass

PR #178, opened 2019-05-11
Revert "Mastodon usernames"

ethanhs:
Reverts ocf/ocflib#172

Because this breaks login for mastodon users

PR #172, opened 2019-05-01
Mastodon usernames

cg505:
don't allow registration of a mastodon username

PR #179, opened 2019-05-13
add validation against mastodon usernames (again)

cg505:
This is my second shot at #172, addressing the issues that caused #178. I also added more tests, including an integration test case that would have caught that error.

### DIFF
--- a/ocflib/account/validators.py
+++ b/ocflib/account/validators.py
@@ -4,6 +4,7 @@ import string
 import sys
 
 import cracklib
+import requests
 
 import ocflib.misc.mail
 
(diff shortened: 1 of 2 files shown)


ethanhs:
@cg505 you can probably add a new keyword argument e.g. `check_mastodon` that is usually `False` then switch that on when it is useful.

dkess:
Can't we just move this logic to the `validate_username` in `creation.py`? That way it won't get called in the login path.

ethanhs:
@dkess yes that is definitely the correct thing to do, missed that was a thing.

PR #180, opened 2019-05-17
Fix the happyhour rule

MCer4294967296:
A more humane way to represent when the double quota days end.
And so that people will have 20 pages today.


dkess:
Wherever this range is define, add a comment specifying that the range is inclusive.

PR #182, opened 2019-06-15
Add printing adjustments

glee-b:
Depended on by ocf/utils#114

Allows staff members to adjust users' paper quotas in a more flexible manner.

We add the following functionality to the `paper` script:
`forward`:  increases a user's daily quota for the day only

This is intended to be used to give users some degree of freedom when printing documents.

For example, if the daily print limit was set at `10` pages and a user wanted to print `12` pages, this would not be possible with the current way things are configured. `forward` provides the way to increase the daily print limit without increasing the amount of pages a user can print in a semester.

### DIFF
--- a/ocflib/printing/ocfprinting.sql
+++ b/ocflib/printing/ocfprinting.sql
@@ -165,4 +165,4 @@ CREATE VIEW `public_jobs` AS
     ORDER BY `day` DESC
 
 GRANT SELECT ON `ocfprinting`.`printed` TO 'anonymous'@'%';
-GRANT SELECT on `ocfprinting`.`public_jobs` TO 'anonymous'@'%';
+GRANT SELECT ON `ocfprinting`.`public_jobs` TO 'anonymous'@'%';
--- a/ocflib/printing/ocfprinting.sql
+++ b/ocflib/printing/ocfprinting.sql
@@ -20,7 +20,7 @@ CREATE TABLE IF NOT EXISTS `adjustments` (
     `id` int NOT NULL AUTO_INCREMENT,
     `user` varchar(255) NOT NULL,
     `time` datetime NOT NULL,
-    `action` varchar(255) NOT NULL,
+    `action` enum('refund', 'forward'),
     `pages` int NOT NULL,
     `staffer` varchar(255) NOT NULL,
     `reason` varchar(510) NOT NULL,
--- a/ocflib/printing/ocfprinting.sql
+++ b/ocflib/printing/ocfprinting.sql
@@ -16,10 +16,11 @@ CREATE TABLE IF NOT EXISTS `jobs` (
     PRIMARY KEY (`id`)
 ) ENGINE=InnoDB;
 
-CREATE TABLE IF NOT EXISTS `refunds` (
+CREATE TABLE IF NOT EXISTS `adjustments` (
     `id` int NOT NULL AUTO_INCREMENT,
     `user` varchar(255) NOT NULL,
     `time` datetime NOT NULL,
+    `action` varchar(255) NOT NULL,
     `pages` int NOT NULL,
     `staffer` varchar(255) NOT NULL,
     `reason` varchar(510) NOT NULL,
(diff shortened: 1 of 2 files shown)


abizer (dismissed):
I'm trusting this works since I'm too tired to parse the SQL right now, lgtm otherwise

glee-:
The associated quota tests need to be updated as well before merging this in

ocflib/printing/quota.py
```diff
@@ -117,6 +118,6 @@ def add_job(c, job):
     c.execute(*_namedtuple_to_query('INSERT INTO jobs ({}) VALUES ({})', job))
 
 
-def add_refund(c, refund):
+def add_adjustment(c, payload):
     """Add a new refund to the database."""
```
dkess: Change this docstring

ocflib/printing/ocfprinting.sql
```diff
@@ -16,18 +16,19 @@ CREATE TABLE IF NOT EXISTS `jobs` (
     PRIMARY KEY (`id`)
 ) ENGINE=InnoDB;
 
-CREATE TABLE IF NOT EXISTS `refunds` (
+CREATE TABLE IF NOT EXISTS `adjustments` (
     `id` int NOT NULL AUTO_INCREMENT,
     `user` varchar(255) NOT NULL,
     `time` datetime NOT NULL,
+    `action` varchar(255) NOT NULL,
```
dkess: This should either be an enum (refund or forward), or you should leave this table alone and make a new `forwards` table. Personally I think the latter is a better option, since every time you use this table you have to check `action` anyways. But either is fine.

dkess (changes requested):
Gonna assume this gets tested a lot before next semester starts. Still, you shouldn't use a varchar where an enum belongs.

ocflib/printing/ocfprinting.sql
glee-: I was initially considering to go with the latter option that you proposed, but I ultimately decided to continue with this current implementation to introduce less complexity for the next person who decides to work on this section of the code.

Performance-wise, I believe that the latter option will be better off as well due to locality, but it is hard to tell for certain because of how SQL operates on the lower layers.

dkess: We don't need to care about performance for this. Why do you think your current solution introduces less complexity? It's slightly less verbose but I would argue needing to put if statements everywhere to check for refund/forward makes things more complicated for anyone working on this.

ocflib/printing/ocfprinting.sql
```diff
@@ -107,9 +137,9 @@ CREATE VIEW `printed_semester` AS
     SELECT
         refunds_semester.user AS user,
         COALESCE(jobs_semester.pages, 0) - COALESCE(refunds_semester.pages, 0) AS semester
-    FROM jobs_semester
-    RIGHT OUTER JOIN refunds_semester
-    ON jobs_semester.user = refunds_semester.user
+    FROM refunds_semester
+    LEFT OUTER JOIN jobs_semester
+    ON refunds_semester.user = jobs_semester.user
```
cg505: why was this change made?

cg505 (commented):
Thank you for working on this!

I'd generally agree with @dkess that I think it would make sense to add a new table. In my opinion that would make it clearer what's going on, (and in a sense, decrease complexity).

ocflib/printing/ocfprinting.sql
abizer: I actually think one table is logically and conceptually simpler. We never interact with this table directly anyways on the read, we do it through the views, which introduce the two table abstraction you're looking for. My conception of this problem is that the difference between refunds and adjustments is whether they permanently increase semesterly quota when increasing daily quota, and given that that is an application-code level concern that just requires a binary switch in the database to decide, it suggests they should continue to live in the same table.

Would switch the column for a enum, or better yet, a boolean, though. eg `is_adjustment` or `is_refund`

PR #184, opened 2019-08-28
Remove "cal" from restricted_usernames

fydai:
People's names can have "cal" in them. For example, reasonable names like "mattmcal" would be rejected for having cal as a substring.

### DIFF
--- a/ocflib/account/creation.py
+++ b/ocflib/account/creation.py
@@ -36,7 +36,7 @@ BAD_WORDS = frozenset((
     'labia', 'muff', 'nigger', 'nigga', 'penis', 'piss', 'prick', 'pube', 'pussy', 'queer', 'scrotum',
     'sex', 'shit', 'slut', 'smegma', 'terrorist', 'twat', 'vagina', 'wank', 'whore'
 ))
-RESTRICTED_WORDS = frozenset(('ocf', 'ucb', 'cal', 'berkeley', 'university'))
+RESTRICTED_WORDS = frozenset(('ocf', 'ucb', 'berkeley', 'university'))
 
 CREATE_PUBLIC_KEY = '''\
 -----BEGIN PUBLIC KEY-----


nattofriends:
Maybe keep it only for groups

2019年8月28日(水) 16:01 fydai <notifications@github.com>:

fydai:
We never have to manually approve these names because there's a huge yellow warning saying "Username <X> contains restricted words". It is very reasonable for names to contain the substring "cal". So I think it's reasonable to default-allow these names.

dkess:
Perhaps an alum from before my time can give more context on why we have this policy in the first place?

kpengboy:
The university doesn't like it when student groups use one of its trademarks (e.g. Cal, Berkeley, ...) IIRC we have traditionally extended this rule to apply to usernames. (Thus an acapella group, for example, can't create an account called `calacapella` or something like that.)

In general, I believe it's wise to respect the university's wishes regarding matters like these. *.berkeley.edu domain names for student groups are, after all, a privilege from the University, one they could take away (and in summer 2013, actually contemplated doing so).

On the other hand, we have persistently had a Scunthorpe Effect problem here. Perhaps we should look into fixing that?

fydai:
All group accounts are manually created (by a staff running approve). As such, I think a suitable compromise would be to remove "cal" from restricted_usernames (allowing users to use a "cal" syllable) and put a (overridable) check into the approve script prompting for confirmation if the username begins with the substring "cal". What do you all think of t his idea?

kpengboy:
IIRC, at least at my time, we had an aversion to individuals picking usernames containing "cal". Perhaps @daradib could shed some light on whether I understood this rule correctly and if it is still useful? If not I would have no objection to restricting only for group accounts.

nattofriends:
Did you read the 8 other comments on this PR?

cg505:
>I think a suitable compromise would be to remove "cal" from restricted_usernames (allowing users to use a "cal" syllable) and put a (overridable) check into the approve script prompting for confirmation if the username begins with the substring "cal".

This makes sense to me. For non-group accounts, things will get caught by the real name checker (a la https://failure.ocf.berkeley.edu/rebuild/2019-11-26#319881).

PR #187, opened 2019-09-22
Rewrite staff_hours functions for new schema from ocf/etc.

redplanks:
Depends on ocf/etc#28

### DIFF
--- a/ocflib/lab/staff_hours.py
+++ b/ocflib/lab/staff_hours.py
@@ -53,6 +53,10 @@ def get_staff_hours():
             continue
 
         for staff_hour in staff_hours['staff-hours'][day]:
+            cancelled = False
+            if 'cancelled' in staff_hour.keys():
+                cancelled = staff_hour['cancelled']
+
             staff_hour_list.append(
                 Hour(
                     day=day,
@@ -64,7 +68,7 @@ def get_staff_hours():
                             position=position(attrs['uid'][0]),
                         ) for attrs in map(user_attrs, staff_hour['staff'])
                     ],
-                    cancelled=staff_hour['cancelled'],
+                    cancelled=cancelled
                 )
             )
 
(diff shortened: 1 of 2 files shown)


ocflib/lab/staff_hours.py
```diff
@@ -30,40 +28,61 @@ def gravatar(self, size=100):
...
-            ],
-            cancelled=hour['cancelled'],
-        ) for hour in staff_hours['staff-hours']
-    ]
+    staff_hour_list = []
+    # If we try iterating through the keys in staff-hours, it's non-deterministic ordering.
+    # Because of the schema, we are guaranteed that these days are always here.
+    for day in ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday']:
+        if not staff_hours['staff-hours'][day]:
+            continue
```
dkess: This line isn't necessary, since each day is guaranteed to be present.

ocflib/lab/staff_hours.py
```diff
@@ -30,40 +28,61 @@ def gravatar(self, size=100):
...
+
+        for staff_hour in staff_hours['staff-hours'][day]:
+            cancelled = False
+            if 'cancelled' in staff_hour.keys():
+                cancelled = staff_hour['cancelled']
```
dkess: This can be shortened to `staff_hour.get('cancelled', False)`

ocflib/lab/staff_hours.py
redplanks: It's an ugly workaround, but if I don't check days in increasing lexical order, then the order of `Hour`s returned is non-determinstic, which breaks the tests. I guess I could use a set in the test?

dkess:
I'm saying you don't get the conditional that checks if each day of the week is present, since it's guaranteed to be present. You can still use this kind of loop.

redplanks:
I need the check since if the day doesn't have any entries under it, the dict entry will be `None`.

PR #199, opened 2019-12-15
use new URL for callink StudentGroupService

cg505:
See the thread that was forwarded to sm@ for more info.
Use https://test-studentgroupservice.sait-west.berkeley.edu/service.asmx
for the test version.


cg505:
and yes I tested it

PR #201, opened 2020-01-30
Remove deprecated is_staff/list_staff functions

jvperrin:
https://github.com/ocf/puppet/pull/837 removed the last uses of these, so they can be removed from ocflib now too.

Unfortunately this can't be merged until the changes from that review are fully active by making all the affected hosts (anthrax, dev-anthrax, and all desktops) use the master branch, or at least updated enough environments that have the changes made to puppet.

### DIFF
--- a/ocflib/account/utils.py
+++ b/ocflib/account/utils.py
@@ -64,26 +64,6 @@ def public_html_path(user):
     return os.path.join(home_dir(user), 'public_html')
 
 
-def is_staff(user, group='ocfstaff'):
-    """Return whether the user is a staff member.
-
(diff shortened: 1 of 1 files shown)


jvperrin:
The tests on this should pass after #202 is merged, currently they're broken for the same reason as https://github.com/ocf/ocfweb/pull/571.

jvperrin:
I've made sure all the desktops have a new enough puppet environment (either production or something else with https://github.com/ocf/puppet/pull/837 included) to apply this, so I've gone ahead and merged it.

PR #209, opened 2020-05-07
set max username length to 16

cg505:
This was causing passwd to fail for usernames longer than 8 chars.

### DIFF
--- a/ocflib/account/utils.py
+++ b/ocflib/account/utils.py
@@ -40,7 +40,7 @@ def extract_username_from_principal(principal):
     'ckuehl'
     """
 
-    REGEX = '^([a-z]{3,8})(/[a-z]*)?@OCF\\.BERKELEY\\.EDU$'
+    REGEX = '^([a-z]{3,16})(/[a-z]*)?@OCF\\.BERKELEY\\.EDU$'
     match = re.match(REGEX, principal)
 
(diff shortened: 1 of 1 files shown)


kpengboy:
Lol, I guess we missed this back in 2015 (or whenever it was).

jvperrin:
Yeah, looks like early 2016 from [rt#4177](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=4177), so it's been more than 4 years now 😅

cg505:
I feel a little scared injecting variables into a regex

nikhiljha:
Side note, I don't actually understand this regex. Shouldn't it be `^([a-z]{3,16})(/[a-z]*)?@OCF\.BERKELEY\.EDU$` instead? I would think that the `\\` is an escaped backslash character, followed by 1 of any non-whitespace character `.`, but when I run both regexes they seem to function as intended.

jvperrin:
It's double escaping, once for python, and once for the dot so it's interpreted as a literal dot. Maybe using a `r` string here would be better to make that less confusing?

cg505:
I'm just going to go ahead and merge this now, the r string would be better but I think we should just get this out. If someone wants to make a PR for that, it would be great but I can't be bothered.

PR #210, opened 2020-05-08
Moved jvperrin's user-email related function here

axmmisaka:
So that it could be reused by other scripts, like `sorry`.

I think pre-commit tests are passed

### DIFF
--- a/ocflib/account/utils.py
+++ b/ocflib/account/utils.py
@@ -2,11 +2,14 @@
 import grp
 import os.path
 import re
+import subprocess
 
 import pexpect
 
 import ocflib.account.validators as validators
 from ocflib.infra.ldap import OCF_LDAP_PEOPLE
+
+LDAP_MAIL_ATTR = 'mail'
 
 
 def password_matches(username, password):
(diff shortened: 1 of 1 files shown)


jvperrin:
Oh, and crucially all of that is a workaround due to `ldap3` not working with authenticated binds on heimdal kerberos and wanting to move off of the last couple remaining python2 scripts (https://github.com/ocf/utils/pull/87), which is to say this is definitely a workaround/hack.

axmmisaka:
Do you want me to put the `kinit` one in ocflib also? There are some prompts inside tho

jvperrin:
I'm honestly not sure. It doesn't feel like it belongs in ocflib due to the prompts that you mentioned (and needing to be user-interactive in general). However, there's some precedent for doing this kind of interactive stuff in ocflib like this: https://github.com/ocf/ocflib/blob/9b031376e2fba1e6edb4e920643dd75cceb29f72/ocflib/misc/shell.py#L33-L67

I'm not convinced we should be following that precedent necessarily, but I do think for this review to go forward in its direction of adding the LDAP-related stuff in ocflib then the kinit stuff should be moved here too alongside it, as it's an (unfortunately) necessary part of it.

axmmisaka:
I'll push more stuff related to that, and see if I can work something out regarding python+kerberos+ldap, thanks for reviews

axmmisaka:
wtf

axmmisaka:
ah, not passing tests seems reasonable as https://github.com/ocf/ocflib/pull/211 isn't merged
i'm creating myself a very big dependency net for PRs like what `apt` would solve

axmmisaka:
Closing https://github.com/ocf/utils/pull/130 https://github.com/ocf/utils/pull/152 https://github.com/ocf/ocflib/pull/210 https://github.com/ocf/utils/pull/153 altogether as I created a looping "merge dependency" almost impossible to review and merge/do CICD; I'll rewrite everything again this semester

PR #213, opened 2020-07-20
Restrict pymysql version to below 0.10.0

ja5087:
this is breaking tests on master due to breaking API changes

imagine releasing a new library version after 2 years and breaking things and no changelog


nikhiljha:
changelog is here(?) btw

https://github.com/PyMySQL/PyMySQL/blob/master/CHANGELOG.md

ja5087:
thanks, I must have missed that

PR #214, opened 2020-07-21
blackify

ja5087:
was testing this out, thoughts? it has weird interactions with flake8 and that's why there are some ignores.


ethanhs:
I personally don't like the black style, e.g. how it splits function definitions.

If we wanted to adopt an autoformatter, I would recommend [`autopep8`](https://pypi.org/project/autopep8/), which uses the same underlying style checker flake8 does.

ja5087:
We already have autopep8. It's a pretty hands off autoformatter actually and time to time we do spend our time commenting on formatting. The chief advantage of black would probably be consistent and unambiguous formatting.

ethanhs:
>It's a pretty hands off autoformatter actually

I kind of like this about autopep8.

>time to time we do spend our time commenting on formatting

Hm, do you have any instances of this in mind? I would imagine this would be useful to report upstream or something.

PR #216, opened 2020-07-26
Remove similarity heuristic

### LINKED ISSUE #136: similarity_heuristic is bad
The code smells and it has a ton of false positives and false negatives. I couldn't even describe to you what the function does without specifying the names of the _multiple_ algorithms involved as well as the order and limits with which it processes match attempts. It would better off returning `random.choice([0, 3])`. This isn't to say that there's a better algorithm that can be used or that the code ought to be cleaned up, though those are probably both true statements, but the fundamental problem with requiring a similar username and real name is that not only is that similarity subjective, but nobody even knows what it means. That makes this not just a code smell, but a UX smell.

- If you want to be real draconian about having usernames be real names, generate the usernames programmatically and don't let the user change it without contacting humans (to deal with the edge case that your programatic solution makes their name sound offensive or something).
- Probably the best thing to do is remove the check and allow some freedom.

bjb568:
Fixes #136.

### DIFF
--- a/tests/account/creation_test.py
+++ b/tests/account/creation_test.py
@@ -211,60 +211,6 @@ class TestCreateDirectories:
                 ]),
             ])
 
-
-class TestUsernameBasedOnRealName:
-
-    @pytest.mark.parametrize('username,realname,success', [
-        ['ckuehl', 'Christopher Kuehl', True],
-        ['ckuehl', 'CHRISTOPHER B KUEHL', True],
-        ['kuehl', 'CHRISTOPHER B KUEHL', True],
-        ['cbk', 'CHRISTOPHER B KUEHL', True],
-
-        ['rejectme', 'Christopher Kuehl', False],
-        ['penguin', 'Christopher Kuehl', False],
-        ['daradib', 'Christopher Kuehl', False],
-    ])
-    @mock.patch('ocflib.account.validators.validate_username')
-    @mock.patch('ocflib.account.search.user_exists', return_value=False)
-    def test_some_names(self, _, __, username, realname, success,):
-        """Test some obviously good and bad usernames."""
-        try:
-            validate_username(username, realname)
-        except ValidationWarning as ex:
-            if success:
-                pytest.fail(
-                    'Received unexpected error: {error}'.format(error=ex),
-                )
-
-    @pytest.mark.parametrize('username', [
-        'shitup',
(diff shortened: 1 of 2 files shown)


tests/account/creation_test.py
```diff
@@ -212,75 +211,6 @@ def test_ensure_web_dir_rename(self):
...
-        """Test some obviously good and bad usernames."""
-        try:
-            validate_username(username, realname)
-        except ValidationWarning as ex:
-            if success:
-                pytest.fail(
-                    'Received unexpected error: {error}'.format(error=ex),
-                )
-
-    @pytest.mark.parametrize('username', [
```
jvperrin, line 239: This and everything below is are not related to the similarity heuristic, so they shouldn't be removed.

jvperrin: Well ok they reference it, but they're testing some other functionality as well

tests/account/creation_test.py
```diff
@@ -212,75 +211,6 @@ def test_ensure_web_dir_rename(self):
...
-            validate_username(username, username)
-
-    def test_error_user_exists(self):
-        """Ensure that we raise an error if the username already exists."""
-        with pytest.raises(ValidationError):
-            validate_username('ckuehl', 'Chris Kuehl')
-
-    @mock.patch('ocflib.account.validators.validate_username')
-    @mock.patch('ocflib.account.search.user_exists', return_value=False)
-    def test_long_names(self, _, __):
```
jvperrin, line 271: This one is still fine to remove though, the permutation issue was only with the similarity heuristic.

bjb568:
This PR is pending BoD approval.

bjb568:
Approved by BoD.

tests/account/creation_test.py
```diff
@@ -474,25 +436,6 @@ def test_valid_request(
             session,
         ) == ([], [])
 
-    @pytest.mark.parametrize('attrs', [
```
ja5087, line 477: This unit tests a lot of other functionality which is not a good sign but probably shouldn't be removed.

ja5087 (changes requested):
otherwise lgtm pending that one test

ja5087 (approved):
Actually, looking at that test again I am not really sure what it's meant to warn against and it shouldn't block this PR.

PR #224, opened 2020-10-09
Allow numbers in usernames

ja5087:
E7 wants a group account with a number in the username and I don't see why not.

### DIFF
--- a/ocflib/account/utils.py
+++ b/ocflib/account/utils.py
@@ -40,7 +40,7 @@ def extract_username_from_principal(principal):
     'ckuehl'
     """
 
-    REGEX = '^([a-z]{3,16})(/[a-z]*)?@OCF\\.BERKELEY\\.EDU$'
+    REGEX = '^([a-z0-9]{3,16})(/[a-z]*)?@OCF\\.BERKELEY\\.EDU$'
     match = re.match(REGEX, principal)
 
     if not match:
--- a/ocflib/account/utils.py
+++ b/ocflib/account/utils.py
@@ -40,7 +40,7 @@ def extract_username_from_principal(principal):
     'ckuehl'
     """
 
-    REGEX = '^([a-z][a-z0-9]{2,16})(/[a-z]*)?@OCF\\.BERKELEY\\.EDU$'
+    REGEX = '^([a-z][a-z0-9]{2,15})(/[a-z]*)?@OCF\\.BERKELEY\\.EDU$'
     match = re.match(REGEX, principal)
 
     if not match:
--- a/ocflib/account/utils.py
+++ b/ocflib/account/utils.py
@@ -40,7 +40,7 @@ def extract_username_from_principal(principal):
     'ckuehl'
     """
 
-    REGEX = '^([a-z0-9]{3,16})(/[a-z]*)?@OCF\\.BERKELEY\\.EDU$'
+    REGEX = '^([a-z][a-z0-9]{2,16})(/[a-z]*)?@OCF\\.BERKELEY\\.EDU$'
     match = re.match(REGEX, principal)
 
     if not match:
--- a/ocflib/account/validators.py
+++ b/ocflib/account/validators.py
@@ -344,8 +344,11 @@ def validate_username(username, check_exists=False):
     if not 3 <= len(username) <= 16:
         raise ValueError('Username must be between 3 and 16 characters.')
 
+    if not username[0].islower():
+        raise ValueError('Username must begin with a lowercase character')
+
     if not all(c.islower() or c.isdigit() for c in username):
-        raise ValueError('Username must be all lowercase and numeric characters.')
+        raise ValueError('Username must be only lowercase and numeric characters.')
 
     if check_exists and not user_exists(username):
         raise ValueError('Username does not exist.')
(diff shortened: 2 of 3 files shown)


kpengboy:
You probably wouldn't want all digit usernames, since those could be confused with UIDs.

ja5087:
I don't think we expose a UID anywhere at the OCF though, but it's fair that just to avoid the possibility in case we ever do to make the first character lowercase alphabetical. As far as I can tell we haven't done anything special that would require more reserved usernames either.

ja5087:
I've changed it so you need to have the first letter be a lowercase character.

ocflib/account/utils.py
```diff
@@ -40,7 +40,7 @@ def extract_username_from_principal(principal):
     'ckuehl'
     """
 
-    REGEX = '^([a-z]{3,16})(/[a-z]*)?@OCF\\.BERKELEY\\.EDU$'
+    REGEX = '^([a-z0-9]{3,16})(/[a-z]*)?@OCF\\.BERKELEY\\.EDU$'
```
PotatoParser, line 43: Not sure why there is a forward slash in the middle of the expression

ja5087: Yup, the idea here is there are multiple types of Kerberos principals at the OCF. For example, I have three principals:

- `jaw/admin@OCF.BERKELEY.EDU` for Kerberos administration
- `jaw/root@OCF.BERKELEY.EDU` for root access
- `jaw@OCF.BERKELEY.EDU` for regular login

https://www.ocf.berkeley.edu/docs/staff/backend/kerberos/

PotatoParser: Ah I see, makes more sense!

tests/account/validators_test.py
```diff
@@ -26,7 +26,6 @@ class TestValidateUsername:
         'Ckuehl',
         'ckuehl!',
         '123123',
```
PotatoParser, line 28: Would this be valid since the regex selects 3-16 digits/lowercase nums

ja5087: It's not valid because the validation logic comes from `validate_username`. You're right that the regex is lax and I don't really know how to feel about it. On the one hand Kerberos principals can be any printable ASCII so there is nothing stopping us from creating `000@OCF.BERKELEY.EDU` for example, and the method's purpose is to extract that out.

ocflib/account/validators.py
```diff
@@ -344,8 +344,11 @@ def validate_username(username, check_exists=False):
     if not 3 <= len(username) <= 16:
         raise ValueError('Username must be between 3 and 16 characters.')
 
-    if not all(c.islower() for c in username):
-        raise ValueError('Username must be all lowercase letters.')
+    if not username[0].islower():
```
kpengboy, line 347: Perhaps while we're here we should also validate that the username contains only ASCII characters?

tests/account/validators_test.py
kpengboy: Is the goal of the regex to handle any arbitrary Kerberos principal, or only those which correspond to valid usernames? I must point out that the regex as written won't match hypothetical Kerberos principals which are longer than 16 characters, have hyphens in them, or the like.

kpengboy:
Here are some additional steps I thought of that may be needed:
 - Do we want to reserve usernames like `adm1n`?
 - We'd need to create `/home/a/a0`, `/home/a/a1`, ..., `/home/z/z9`.

tests/account/validators_test.py
ja5087: That's fair. I think making it tighter and in line with what we allow will be better than in this weird state of not doing either.

ja5087:
> 
> 
> Here are some additional steps I thought of that may be needed:
> 
>     * Do we want to reserve usernames like `adm1n`?
> 
>     * We'd need to create `/home/a/a0`, `/home/a/a1`, ..., `/home/z/z9`.

Considering we don't reserve things like `admin` either and it's probably impossible to maintain a list of all things that sound important, this might be a wontfix. As for the second point, I think this is already ensured by https://github.com/ocf/ocflib/blob/309680aaa2b9e3e6315ad52d0f9118b0a9482acd/ocflib/account/utils.py#L52 and when I tested this function to create `e7staff` it worked fine.

kpengboy:
We do reserve `admin` FWIW: https://github.com/ocf/ocflib/blob/309680aaa2b9e3e6315ad52d0f9118b0a9482acd/ocflib/account/validators.py#L18.
But agree we can consider reserving such leetspeak usernames in a different change.

ja5087:
Oh huh I stand corrected, thanks.

ethanhs:
I think talking to @nikhiljha, we're a bit hesitant to merge this, as people could use numbers to make their usernames a bit more obfuscated, e.g. `m1337ax` or something. I think just typing out the numbers (.e.g `7` is `seven`) is what we have done in the past.

axmmisaka:
iirc we do have requirements that username must somewhat match real name and it's automatically enforced somewhere - could possibly be useful to prevent somebody using "adm1n" (the one who exploit it must has a name like Adam Norton or something...)

PR #228, opened 2021-01-26
Try unpinning pymysql

ethanhs:


ethanhs:
Ah, new pymysql is incompatible with 3.5, never mind then...

PR #232, opened 2021-10-02
Added get newest RT ticket

lancemathias:
New method to get latest RT ticket using REST API. Pushed to master to test since local tests aren't working.

### DIFF
--- a/ocflib/infra/rt.py
+++ b/ocflib/infra/rt.py
@@ -37,6 +37,28 @@ class RtTicket(namedtuple('RtTicket', ('number', 'owner', 'subject', 'queue', 's
         )
 
     @classmethod
+    def get_newest(cls, connection, queue):
+        """Returns the newest created RT ticket in the given queue"""
+        resp = connection.get("https://rt.ocf.berkeley.edu/REST/1.0/search/ticket?query=Queue='{}'&orderby=-Created".format(queue))
+        assert resp.status_code == 200, resp.status_code
+        assert '200 Ok' in resp.text
+
+        lines = resp.text.splitlines()
+
+        def find(header):
+            for line in lines:
+                if line.startswith(header + ': '):
+                    return line.split(': ', 1)[1]
+
+        return cls(
+            number=num,
+            owner=find('Owner'),
+            subject=find('Subject'),
+            queue=find('Queue'),
+            status=find('Status'),
+        )
+
+    @classmethod
     def create(cls, connection, queue, requestor, subject, text, **kwargs):
         """Create an RT ticket and returns an instance of the result"""
         # RT prefixes multiline strings by a blank space


ethanhs:
(Closed and reopened, which seems to have fixed CI not getting triggered initially)

ocflib/infra/rt.py
```diff
@@ -36,6 +36,28 @@ def find(header):
             status=find('Status'),
         )
 
+    @classmethod
+    def get_newest(cls, connection, queue):
+        """Returns the newest created RT ticket in the given queue"""
+        resp = connection.get("https://rt.ocf.berkeley.edu/REST/1.0/search/ticket?query=Queue='{}'&orderby=-Created".format(queue))
```
axmmisaka, line 42: Not sure if this would work since I can't seem to find relevant stuff in the docs (https://rt-wiki.bestpractical.com/wiki/REST#Ticket_Search), but I think it might return a bunch of tickets instead of one?

ocflib/infra/rt.py
```diff
@@ -36,6 +36,28 @@ def find(header):
...
+
+        lines = resp.text.splitlines()
+
+        def find(header):
+            for line in lines:
+                if line.startswith(header + ': '):
+                    return line.split(': ', 1)[1]
+
+        return cls(
+            number=num,
```
axmmisaka, line 54: Probably need to do something to find `id: ticket/<ticket-id>`;

ocflib/infra/rt.py
```diff
@@ -36,6 +36,28 @@ def find(header):
...
+    @classmethod
+    def get_newest(cls, connection, queue):
+        """Returns the newest created RT ticket in the given queue"""
+        resp = connection.get("https://rt.ocf.berkeley.edu/REST/1.0/search/ticket?query=Queue='{}'&orderby=-Created".format(queue))
+        assert resp.status_code == 200, resp.status_code
+        assert '200 Ok' in resp.text
+
+        lines = resp.text.splitlines()
+
+        def find(header):
```
axmmisaka, line 48: idk this is legacy, it's elegant but comparatively inefficient i guess (need to do 4 passes instead of 1)

axmmisaka (changes requested):
thanks for the good work, i'm still figuring out how to access rt api with my acct

ocflib/infra/rt.py
axmmisaka:
```
Invalid query: 'Wrong query, expecting a VALUE in 'Queue=>help<--here' at /opt/rt5/sbin/../lib/RT/Tickets.pm line 3036, <DATA> line 1662.

Stack:
  [/opt/rt5/sbin/../lib/RT/Tickets.pm:3036]
  [/opt/rt5/sbin/../lib/RT/Tickets.pm:3248]
  [/opt/rt5/share/html/REST/1.0/search/dhandler:120]
  [/opt/rt5/share/html/REST/1.0/autohandler:54]
  [/opt/rt5/sbin/../lib/RT/Interface/Web.pm:710]
  [/opt/rt5/sbin/../lib/RT/Interface/Web.pm:389]
  [/opt/rt5/share/html/autohandler:53]
'.
```

axmmisaka:
https://metacpan.org/pod/RT::Extension::REST2#Cookie-Auth
There's some discussion in #rebuild regarding using REST 2.0 api. The issue is for REST 2.0 it appears that the docs suggest against using session-token to authenticate, so every time auth token/password is needed in the header if I understand it correctly?
Otherwise use `resp = s.get('https://rt.ocf.berkeley.edu/REST/2.0/tickets?query=Queue="help"&orderby=Created&order=DESC', headers={"Authorization": "token dskljvelkajdhvluierwbvlijaebvleisuadghea"})` json parse `resp.text`, and pick the first in `items`.

PR #233, opened 2021-10-09
Update validators.py

singingtelegram:
Add reserved word: merch (merch@obe is the RT queue for the merch store)


emmatyping (commented):
Looks like we should bump the known UID (which you can do here or another PR if you want) and you need to add to the length of the list...

fawaf:
woah, ocf will have a merch store soon?

PR #235, opened 2021-10-21
typo: replace underscore with hyphen

ben9583:
Accidentally typed `meeting_hours` instead of `meeting-hours`, which is the correct value in the yaml file on etc.


ben9583:
Caught this when making some changes on ocfweb.

PR #238, opened 2021-11-13
Meeting hours fixes + tests

ben9583:
Fixes all the issues with meeting hours and adds tests for related methods as well.

### DIFF
--- a/ocflib/org/meeting_hours.py
+++ b/ocflib/org/meeting_hours.py
@@ -104,6 +104,25 @@ def _time_to_range(hours):
     return (start_time, end_time)
 
 
+def _iso_weekday_to_str(num):
+    if(num == 1):
+        return "Monday"
+    elif(num == 2):
+        return "Tuesday"
+    elif(num == 3):
+        return "Wednesday"
+    elif(num == 4):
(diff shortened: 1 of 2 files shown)


ocflib/org/meeting_hours.py
```diff
@@ -86,27 +86,45 @@ def _time_to_range(hours):
...
     return (start_time, end_time)
 
 
-def _get_next_meeting():
-    today = date.today()
-    now = localtime().tm_hour * 60 + localtime().tm_min
+def _iso_weekday_to_str(num):
+    if(num == 1):
+        return 'Monday'
+    elif(num == 2):
```
Kalissaac, line 110: One small thing is that it seems the rest of the codebase doesn't use parentheses around if conditionals, the linter should fix this automatically but I can't seem to get it to run right now

Kalissaac (approved):

ocflib/org/meeting_hours.py
ben9583: Ah yea forgot about that python syntax quirk; I'll get that fixed

ben9583:
Thank you for the review!

PR #239, opened 2021-12-03
ignore UIDs created as a result of CTF mishap

ethanwu10:
During the CTF, a user was created with UID 13371337 for a challenge, which resulted in new users being created with UIDs > 13370000. Ignore all such users when calculating the next UID to generate (the current users in that UID range will keep their UIDs).

### DIFF
--- a/ocflib/account/creation.py
+++ b/ocflib/account/creation.py
@@ -54,6 +54,11 @@ RESERVED_UID_RANGES = [
     (61184, 65535),
 ]
 
+IGNORED_UID_RANGES = [
+    # Incidient with OCF CTF resulted in users being created in this range
+    (13371337, 13371800),
+]
+
 
 def _get_first_available_uid(known_uid=_KNOWN_UID):
     """Return the first available UID number.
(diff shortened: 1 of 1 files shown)


kpengboy:
- Are we doing anything to prevent this from happening in the future?
 - Do these high UID accounts have proper quotas set up?
 - Have you considered the implications if we ever want to set up subuid for per-user Docker or the like?

fawaf:
ocf had a ctf? as in capture the flag?

ethanwu10:
> * Are we doing anything to prevent this from happening again in the future?

To be entirely honest this PR is just to stop new accounts from continuing to be registered with high UIDs, and the only way I really see to fix the problem is to either make the UID selection logic more robust (which will take a while), carve out and document a block of UIDs for future "system" accounts, or both.

>     * Do these high UID accounts have proper quotas set up?

As far as I can tell yes; we already have many UIDs >16bits, so there shouldn't:tm: be any problems here

>     * Have you considered the implications if we ever want to set up subuid for per-user Docker or the like?

I'm not sure how that's relevant here since this just prevents the UID selection logic from picking these UIDs (unless you're referring to the high UIDs themselves, which is a valid concern)

> If we can migrate these users to lower UIDs, maybe that would be more robust in the longer run...

I definitely agree, however that carries some risk while merging this ASAP would at least decrease the number of users we'd need to migrate (though we'd probably need to also go carve out another block in a lower range to migrate *to*)

kpengboy:
Personally, I would prefer:
 - If we anticipate that such test accounts might be created again in the future, we might want to blanket-ignore all UIDs under a certain fixed high number (e.g. above 10 million).
 - IMO we should still try to migrate the created accounts down to lower numbers if possible.

singingtelegram (approved):
lgtm

PR #240, opened 2021-12-04
Update Jenkinsfile

singingtelegram:
Build for bullseye

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 packagePipeline(
     coverallsToken: 'coveralls_ocflib_token',
     pypiRelease: true,
-    dists: ['stretch', 'buster'],
+    dists: ['stretch', 'buster', 'bullseye'],
 )
 
 // vim: ft=groovy


ethanhs:
We need a base image before this will work

singingtelegram:
Built, see ocf/dockers.

ethanwu10 (changes requested):
We should resolve [build failures](https://jenkins.ocf.berkeley.edu/blue/organizations/jenkins/ocf%2Fdockers/detail/master/514/pipeline#step-35-log-1843) in ocf/dockers first

ethanwu10:
Turns out the build failures are from [autoclean-docker](https://github.com/ocf/puppet/blob/master/modules/ocf_jenkins/files/autoclean-docker) deleting the temp image while the job was in the middle of building since reaper had <25% free disk on `/`. I'll open a new issue for that on puppet, but this PR seems good to me.

ethanwu10 (approved):

PR #246, opened 2022-07-10
Add support for null/closed hours

Boomaa23:
Adds support for closing the lab for the entire day regularly (such as the current situation with MLK being closed on Sundays). More info: https://github.com/ocf/etc/pull/250#issuecomment-903232452

To close for a day, edit [etc/hours.yaml](https://github.com/ocf/etc/blob/master/configs/hours.yaml) to something like the following (this example would close Sunday):
```yaml
regular:
  ...
  Satuday:
    - ['11:00', '18:00']
  Sunday: null
```

The result is something that looks like the following (note the "Closed All Day" days)
hours

### DIFF
--- a/ocflib/account/creation.py
+++ b/ocflib/account/creation.py
@@ -24,7 +24,7 @@ from ocflib.misc.validators import valid_email
 from ocflib.printing.quota import SEMESTERLY_QUOTA
 
 
-_KNOWN_UID = 74209
+_KNOWN_UID = 78418
 BAD_WORDS = frozenset((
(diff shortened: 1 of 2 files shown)


Kalissaac (commented):
Should we bump the `num_uids` constant before we merge to fix CI?

Boomaa23:
For reference, the following works without changing ocflib
```yaml
Sunday: []
```
and if ocflib is changed, these would *also* work
```yaml
Sunday:
Sunday: null
```
This does not work in any circumstances (unless we change it to also cover `[[]]`)
```yaml
Sunday:
  - []
```

nikhiljha (approved):

PR #257, opened 2023-10-31
Create ocflib Tools for Announcements

### LINKED ISSUE #49: Create ocflib Tools for Announcements
We'll use `ocflib` to get information on announcements, which the API will call as an interface between the user and our backend. `ocflib` should support the following specifications:

- [ ] Get metadata for a post
- [ ] Get content for a post
- [ ] Get the IDs for the last `num` posts

These implementations should make use of the GitHub API to find and retrieve the file contents of a given announcement. They should also make use of caching so we don't have to do this often (see other ocflib projects for specific details).

vicshi06:
https://github.com/ocf/api/issues/49

### DIFF
--- a/ocflib/lab/announcements.py
+++ b/ocflib/lab/announcements.py
@@ -37,9 +37,13 @@ def get_announcement(id: str) -> str:
     return posts.text
 
 
-def get_metadata(id: str) -> Metadata:
+def get_id(post: str) -> str:
+    """Get announcement id"""
+    return post["name"][:-3]
+
+
+def get_metadata(post: str) -> Metadata:
     """Get the metadata from one announcement"""
-    post = get_announcement(id)
     meta_dict = safe_load(post.split("---")[1])
 
     metadata = Metadata(
@@ -52,19 +56,16 @@ def get_metadata(id: str) -> Metadata:
     return metadata
 
 
-def get_all_ids():
-    """Get all announcement ids"""
+def get_announcements(num: int) -> list[str]:
+    """Get the last num announcements"""
     posts = get_all_announcements()
-    return [post["name"][:-3] for post in posts]
 
-
-def get_ids(num: int):
-    """Get the last num announcement ids"""
-    posts = get_all_announcements()
-    return [post["name"][:-3] for post in posts[:num]]
+    # get the last num posts in reverse order
+    return [get_announcement(get_id(post)) for post in posts[-1 : -num - 1 : -1]]
 
 
 # print(get_all_announcements())
 # print(get_announcement("2002-01-01-00"))
-# print(get_metadata("2002-01-01-00"))
+# print(get_metadata(get_announcement("2002-01-01-00")))
 # print(get_all_ids())
+# print(get_announcements(2))
--- a/ocflib/lab/announcements.py
+++ b/ocflib/lab/announcements.py
@@ -1,39 +1,70 @@
 """Announcements handling"""
-import requests
-import json
+from collections import namedtuple
+from requests import get
+
+from yaml import safe_load
+from typing import NamedTuple
 
 ANNOUNCEMENTS_REPO = (
     "https://api.github.com/repos/ocf/announcements/contents/announcements/{id}"
 )
 
+# TODO: cache
+# TODO: request session for better efficiency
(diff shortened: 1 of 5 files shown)


vicshi06:
@ben9583 Can you take a look at the first draft? I am able to use the GitHub API to get the content and the metadata of one specific announcement based on the ID. I will work on the third feature later. Here is the output for the functions. 

## Content 
```bash
---
title: "testing"
date: 2002-01-00
author: Victor
tags: [test, test2, test3]
summary: this is a test announcement 
---

# Test Announcement

## Summary

- this is a test announcement.

## Detail

- detail of the test annoucement.

## Contacts

- Name: Victor
- Email: contact@berkeley.edu
- Phone: (123)456-7890

```

## Metadata
```bash
title: "testing"
date: 2002-01-00
author: Victor
tags: [test, test2, test3]
summary: this is a test announcement 
```

vicshi06:
@ben9583 Hi Ben. Can you take a look and make sure the function behaviors are good? I will implement cache and make the code more efficient later.

ocflib/lab/announcements.py
```diff
@@ -0,0 +1,71 @@
...
+
+def get_announcements(num: int) -> list[str]:
+    """Get the last num announcements"""
+    posts = get_all_announcements()
+
+    # get the last num posts in reverse order
+    return [get_announcement(get_id(post)) for post in posts[-1 : -num - 1 : -1]]
+
+
+# print(get_all_announcements())
```
ben9583, line 67: Let's get rid of these debug statements for the pr

ocflib/lab/announcements.py
```diff
@@ -0,0 +1,71 @@
...
+        headers={"Accept": "application/vnd.github.raw"},
+    )
+    posts.raise_for_status()
+
+    return posts.text
+
+
+def get_id(post: str) -> str:
+    """Get announcement id"""
+    return post["name"][:-3]
```
ben9583, line 42: Can you enumerate more why this works? Just so that way other people know what's going on here.

ben9583 (changes requested):

ocflib/lab/announcements.py
vicshi06: Since the name of the announcements is `YYYY-MM-DD.md`, this is to grab everything except for the last 3 characters. I can put more comments in the code as well.

ben9583: Yea I remember, but I think comments in code would be good.

ben9583:
As for why the tests are failing, one of them is a regular housekeeping task (don't worry about it), and the other two are... from my code from 2 years ago. I'll take a look at why they're failing now.

ben9583:
Ok I'm blaming the issues with mine on daylight savings happening that day. There only seems to be one commit in which it fails and that doesn't include the most recent one

vicshi06:
Ok is there anything I need to do to fix the checks?

ben9583:
Will take a more thorough look at this tomorrow, but it seems the [tests on Jenkins](https://jenkins.ocf.berkeley.edu/blue/organizations/jenkins/ocf%2Focflib/detail/PR-257/7/pipeline/30/#step-33-log-1) are not working properly since the recent changes

ben9583:
This is looking pretty much all good, just a couple notes

- The post IDs should be cached too (in addition to the text which you've done), so that way we don't have to call the GitHub API to scan all of the announcements so often
- There should be `get_last_n_announcements` that just returns the IDs and metadata (using caching), we don't need to get all the last `n` announcement full text contents

Let me know if you have questions of what I mean by this cause I may have worded this a bit weird. I can explain more on specifics

vicshi06:
I have changed the structure a little. The script now checks for updated cache whenever `get_last_n_announcements` and `get_last_n_announcements_text` are called, so the cache checking behavior is more consistent. 

I have also updated the `README` so people would know how to test individual test file. 

Let me know if you have any thoughts! The script should be good to go. I haven't added the tests for the cache since I have not yet thought of a good way to test it. It might make more sense to actually call it from the front end to make sure its behavior is good. But it is not super complex so we don't necessarily need super complex tests.

PR #258, opened 2024-03-15
Use poetry for dependencies & build

oliver-ni:
Shifting to use Poetry to manage dependencies, build, and local environments. Replaces old venv-update workflow. Makefile has been updated accordingly and things like Jenkins should work smoothly with the new setup.

Also adds a nixpkgs overlay that overrides the nixpkgs Python to add the ocflib package.


Makefile
```diff
@@ -1,40 +1,36 @@
-venv: setup.py requirements-dev.txt
-	vendor/venv-update \
-		venv= $@ -ppython3 \
-		install= -r requirements-dev.txt -e .
+venv: poetry install
 
 .PHONY: install-hooks
 install-hooks: venv
```
nikhiljha, line 4: This doesn't appear to work...

```
❯ make install-hooks
make: *** No rule to make target `poetry', needed by `venv'.  Stop.
```

nikhiljha (changes requested):
looks good in general but we should fix pre-commit

Makefile
oliver-ni: whoops I need a newline there

nikhiljha (approved):

PR #261, opened 2024-04-20
Add secondary IPv4 subnet to list of OCF subnets

Kalissaac:
This subnet is intended for DeCal VMs but might also be used for other hosts.

Right now we have a desktop assigned to an IP in this subnet and ocfweb is freaking out because the IP isn't officially in our subnet according to ocflib.


oliver-ni:
Whoops. I clobbered some desktops' IPs with the new ones that came in, so temporarily assigned those to the other subnet (it probably doesn't even work)

oliver-ni (approved):

PR #271, opened 2025-03-15
fix: allow get_quota to return negative quotas

xa1on:
revoking a large number of pages results in get_quota returning 0 despite the true value being negative. this avoids confusion by allowing users to see how much paper debt they are in.


laksith19 (approved):
lgtm!

PR #287, opened 2026-05-06
add fishpaper and printer status query

BNH440:
adds fishpaper to list of printers and adds a way to query the printer status string via snmp (gets data like if printer is out of paper)

### DIFF
--- a/ocflib/printing/printers.py
+++ b/ocflib/printing/printers.py
@@ -13,11 +13,21 @@ OID_MAINTKIT_CUR = '1.3.6.1.2.1.43.11.1.1.9.1.2'
 
 OID_LIFETIME_PAGES_PRINTED = '1.3.6.1.2.1.43.10.2.1.4.1.1'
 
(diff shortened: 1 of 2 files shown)


jaysa68 (dismissed):
rebase on main and ensure tests pass + it builds, then you are good to go!

BNH440:
going to add paper tray status too

ocflib/printing/printers.py
```diff
@@ -3,7 +3,7 @@
 
 import puresnmp
 
-PRINTERS = ['logjam', 'pagefault', 'papercut']
+PRINTERS = ['logjam', 'pagefault', 'papercut', 'fishpaper']
```
24apricots, line 6: can we query ldap and get printers in prod env

24apricots (approved):

ocflib/printing/printers.py
BNH440: that would be nice but probably not realistic right now as we'd also need to tell the difference between hp printers and not

PR #298, opened 2026-06-10
Change group utils to pull from ldap directly

vqbc:
probably merge this sooner rather than later...we'd have to rebuild ocfweb manually for the fix to take effect im assuming

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -92,7 +92,7 @@ in
 
 buildPythonPackage {
   pname = "ocflib";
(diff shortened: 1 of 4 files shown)


BNH440 (commented):
we should update the version in default.nix as well, although it appears this has been forgotten for some time now (like setup.py, thanks for bumping that one)

looks good otherwise

BNH440 (approved):

PR #299, opened 2026-06-10
fix expected kadmin path in test

kafu-densha:
set the expected kadmin path to the location on nixos which fixes the test failures when testing on `koi`

this may be a bad idea if the runner isn't in a nix environment as it would cause CI to fail


BNH440:
superseded by #302

PR #300, opened 2026-07-14
use ldap3 from Nixpkgs

dotlambda:
Otherwise using `python3.withPackages (ps: [ ps.ldap3 ps.ocflib ])` results in
```
pkgs.buildEnv error: two given paths contain a conflicting subpath:
  `/nix/store/c3592ilp83az3vyrcgbmxv6iiqxbz9l8-python3.14-ldap3-2.9.1/lib/python3.14/site-packages/ldap3/utils/ciDict.py' and
  `/nix/store/cml2p3mb2jnws0yk9ziwyd67m3gdq4c9-python3.14-ldap3-2.9.1/lib/python3.14/site-packages/ldap3/utils/ciDict.py'
hint: this may be caused by two different versions of the same package in buildEnv's `paths` parameter
hint: `pkgs.nix-diff` can be used to compare derivations
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -14,6 +14,7 @@
 , attrs
 , dnspython
 , jinja2
+, ldap3
 , passlib
 , pexpect
 , pycryptodome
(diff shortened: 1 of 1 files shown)


sophiebsw (approved):
lgtm

sophiebsw:
ah right ci is mega messed up, i'm going to override to merge because this is a nix change and the flake outputs is separate from the ci

PR #305, opened 2026-09-17
printing: add fishpaper to color queue list

kafu-densha:
I'm changing the server-side name of the color printer from OCF-Color to fishpaper so this db change is required to ensure color prints continue counting towards color quota

this can be merged immediately as it just adds another printer that counts as color but does not affect current operations even if ocf/nix#422 isn't merged yet


kafu-densha:
fyi in my experience this db schema never applies automatically so I will apply it manually to the db once merged (as I did last time)

sophiebsw (approved):

PR #307, opened 2026-09-23
add privileged ucb ldap bind

skybound0:
replace current anonymous ucb ldap bind


kafu-densha (approved):

kafu-densha (changes requested):
disable failing tests and create an issue, or we can try to fix the tests now

kafu-densha (approved):

